PR #8, opened 2015-09-12
Use the samba-provided CHECK_CFG method in configuration tests

ita1024:
This is for forward compatibility with waf 1.8. All other tests
use CHECK_CFG, but check_cfg was re-introduced for some reason.


abartlet:
Merged as 8262ecb98884cb71a59c6a18252aa632b8cbcc35

PR #9, opened 2015-09-13
Client timeout

Gazzonyx:
Fix client socket timeout to be consistent between source3 and source4.  This changes source4's socket timeout from 15 seconds to the old source3 value of 30 seconds.  

Could I get two Team members to sign off on this, please?


Gazzonyx:
*sigh* I didn't want to include my pull of master, but it doesn't change this commit functionally.

PR #11, opened 2015-09-17
Client timeout 2

Gazzonyx:
Respin of the client timeout patch.

Off topic :
Somehow I always get myself in a situation with Git where I can't just cherry pick a single commit.  Disregard the pull from master, I was syncing up and now am forever stuck with a no-op commit in my branches.

### DIFF
--- a/lib/socket/socket_common.h
+++ b/lib/socket/socket_common.h
@@ -1,0 +1,33 @@
+/*
+Unix SMB/CIFS implementation.
+
+Common socket definitions and default values.
+
+Copyright (C) Andrew Tridgell 1992-1998
+Copyright (C) Luke Kenneth Casson Leighton 1996-1998
+Copyright (C) Jeremy Allison 1998
+
+This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


sathieu:
Maybe #9 should be closed then?

Gazzonyx:
Yeah.  Good call.  Closing #9.

Gazzonyx:
I'm going to close this PR due to bit rot and rebase.  I had an interesting conversation with Chris Hertel and Jose Rivera last night in Minneapolis; I need to read up on SMB internals a bit before I'm comfortable with this patch.  Chris described how the timeout mechanism on the Windows side works a bit, from legacy to current as he could best remember and it's frightful.  

Which brings me to my philosophical question for anyone that cares to chime in; do we care if the Samba client side truly emulates a Windows client or only that it interfaces well enough and doesn't do anything "clever" when talking to non-Windows hosts?  To be clear, this is mostly academic since more modern SMB dialects aren't nearly as scary, as I understand it, on both ends of the wire.  This is more about future changes than this particular patch set.

PR #14, opened 2015-09-24
CTDB: Use socket_wrapper only with selftest

sathieu:


sathieu:
Same as c0a463d94abb5c50eaca7b1d402c979684f96a97

gladiac:
Please add your Signed-off: to the patch ...

obnoxxx:
The patch itself looks good to me, but it seems a rebase is needed with conflict resolution.

amitay:
This patch has already been pushed.

obnoxxx:
ok ... communication on too many sides ... :-)

PR #15, opened 2015-09-28
build:wafsamba: Set the default installation prefix for Waf 1.8

ita1024:
These changes enable the default installation prefix settings to
take effect in both Waf 1.5 and 1.8 with no additional code changes.


obnoxxx:
pushed this to autobuild.
still not 100% certain how this thing works in relation to samba-technical :-)

abartlet:
Merged as 71a168eb2bc9df448bab73ae55e2cd8b514b03cd

PR #17, opened 2015-10-07
Add --set-nt-hash parameter to pdbedit to update user password from nt-hash hexstring.

bud4:
I modify pdbedit to   set password from  nt-hash to an existing user.This 
is useful to take in sync password  (for example from OpenLdap) .It is 
also updated password-history.
pbedit  with -vw (verbose & smbpasswd-style )  shows the password hashes 
in the verbose output.

Signed-off-by: Alberto Maria Fiaschi <alberto.fiaschi@estar.toscana.it>

### DIFF
--- a/docs-xml/manpages/pdbedit.8.xml
+++ b/docs-xml/manpages/pdbedit.8.xml
@@ -43,6 +43,7 @@
 		<arg choice="opt">-r</arg>
 		<arg choice="opt">-s configfile</arg>
 		<arg choice="opt">-S script</arg>
+		<arg choice="opt">--set-nt-hash</arg>
 		<arg choice="opt">-t</arg>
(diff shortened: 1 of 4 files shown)


bud4:
Code style fixed.

abartlet:
If it isn't too much trouble, can you squash the two patches into one?

Also, we really should have tests for this.  If you feel inspired, could you find a suitable test to extend to check this code still works (it is OK to do this in a source3 test if writing one to hook into the AD DC is too hard)

bud4:
Rebase last two commit.
Now ,I'm full time on other activity . I hope to try to write some test next week, but I'm not sure.

abartlet:
Merged as 0548fc52f99f3f222b4e58f587dd07f92be6707c

PR #19, opened 2015-10-16
build: Build *_wrapper without -DNDEBUG for in-tree use

abartlet:
These binaires are not installed, so are only used in make test,
and there we need debug output.


abartlet:
Merged as 2bad085305f8ddad5e46a567c63a244362699187

PR #20, opened 2015-10-19
Simple samba fixes

abartlet:
A small set of fixes to Samba that should be 'obviously correct' enough for an easy review.


abartlet:
Merged as f00dd4a8d13bb6806ac2b123a99ad9db8a8a2482

PR #21, opened 2015-10-19
ldb: Fix python bindings to accept a string as a DN

abartlet:
This fixes add_base(), add_child() and is_child_of().

This removes a toally incorrect cast of struct ldb_dn to struct ldb_context.

A helper routine is used instead


abartlet:
Merged as 1c02f2801ecd05df23730eea2b6e162f4d5186f1

PR #22, opened 2015-10-19
Domain clone

abartlet:
samba-tool: Add new command 'samba-tool drs clone-dc-database'
    
    This command makes a clone of an existing AD Domain, but does not
    join the domain.  This allows us to test if the join would work
    without adding objects to the target DC.
    
    The server password will need to be reset for the clone to
    be any use, see the source4/scripting/devel/chgtdcpass
    
    (Based on patches written with Garming Sam)
    
    Andrew Bartlett


abartlet:
Merged as ffe8090c898ac649b132fbfc7d79965dd98ff6c1

PR #25, opened 2015-10-20
pydsdb: Also accept ldb.MessageElement values to dsdb routines

abartlet:
This shows the correct way to accept a value that may be a list of strings
or a proper ldb.MessageElement.

Andrew Bartlett


abartlet:
Merged as b48776d78b446ad4abd4a6bc2ba6b488a29b11d2

PR #26, opened 2015-10-20
repl_meta_data fixes

abartlet:
While developing the demote-dc branch adding better support for 'samba-tool domain demote --remove-other-dead-server' command I tested against a clone of a Windows 2012R2 server.  Removing an RODC triggered a failure due to the binary linked attribute msDS-RevealedUsers


abartlet:
Partially merged as 6cd8e79257a653e5df79c6bebb8689c0203a4f48 but the patch "repl_meta_data: Remove the correct linked attribute forward link" breaks 'make test' and needs more investigation.

PR #27, opened 2015-10-24
build:wafsamba: dead code removal in gettext detection

ita1024:
Removed the --with-gettext option as it has no effect and the code is broken and misleading. Should such options added in the future, use the samba functions such as ADD_CFLAGS.

Reviewed-by: <your name here>
Reviewed-by: <your name here>

### DIFF
--- a/lib/replace/wscript
+++ b/lib/replace/wscript
@@ -361,13 +361,6 @@ removeea setea
     # try to find libintl (if --without-gettext is not given)
     conf.env.intl_libs=''
     if not Options.options.disable_gettext:
-        # any extra path given to look at?
-        if not Options.options.gettext_location == 'None':
-           conf.env['CFLAGS'].extend(["-I%s" % Options.options.gettext_location]);
-           conf.env['LDFLAGS'].extend(["-L%s" % Options.options.gettext_location]);
-        else:
-           conf.env['CFLAGS'].extend(["-I/usr/local"]);
-           conf.env['LDFLAGS'].extend(["-L/usr/local"]);
         conf.CHECK_HEADERS('libintl.h')
         conf.CHECK_LIB('intl')
         conf.CHECK_DECLS('dgettext gettext bindtextdomain textdomain bind_textdomain_codeset', headers="libintl.h")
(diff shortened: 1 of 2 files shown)


lib/replace/wscript
```diff
@@ -362,12 +362,9 @@ removeea setea
     conf.env.intl_libs=''
     if not Options.options.disable_gettext:
         # any extra path given to look at?
-        if not Options.options.gettext_location == 'None':
-           conf.env['CFLAGS'].extend(["-I%s" % Options.options.gettext_location]);
```
urisimchoni: Wouldn't conf.ADD_EXTRA_INCLUDES be more suitable? No a waf expert but it always seems more reasonable to me to separate compiler translation flags from pre-processor flags.
If that's not suitable for some reason, I see we also have ADD_CFLAGS...

ita1024:
I see three possibilities here:

1. Removing the whole "gettext_location" option may be the best course of action since the whole thing is broken. But are we allowed to remove command-line options? I can update the pull request if you think that this is better.

2. Since the code had no effect and gettext_location seems unused, the changes suggested would not make the situation any worse.

3. Using ADD_CFLAGS may be best, but I do not understand how that works either. Also the overall logic of adding the same path to link flags and to compilation flags also incorrect. If you are looking for an overall refactoring of this I will cancel the PR.

urisimchoni:
Let's see...
Seems that:
1. Beside gettext, the only other places where a configurable location is allowed are:
a. ceph/gpfs. Ceph and gpfs are not ordinary system libraries so let's put them aside for the moment.
b. --with-libiconv= - that seems to be a noop, but in a differet way (storing a value in the option and not doing anything with it).
So, unless I'm missing something, the feature of defining a custom dir is not so widespread.

2. The implementation of this feature is quite broken even if the compiler flags do get fixed, because even if you have it in some custom dir, the headers and libs would not normally be under the same dir, they will be in /foo/include, /foo/lib, etc.

3. None complained (?) about this being broken

If all the above is correct, I'm for removing this.

ita1024:
No, no one complained. I have verified that running `configure.developer --with-gettext=/xyz` has no effect whatsoever.

The `--with-gettext` option has been removed in the updated pull request.

abartlet:
Merged as 84b7a9f939f43baf33de4384a8c06687dd4e9c2b

PR #31, opened 2015-10-31
build:wafsamba: refactored the gcc-based dependency calculation

ita1024:
* The option --enable-gccdeps had no effect (store_true),
  so it has been changed to --disable-gccdeps
* The duplicate gccdeps file provided in the Samba tree has
  been removed. It was already present as a Waf extension under
  the third_party directory.
* The module gccdeps has been backported from waf 1.8. Besides
  a few bug fixes, this new version will raise errors in case
  dependencies cannot be found
* The redundant tests present in the samba scripts have been removed
* The gccdeps extension will use the -MMD flag instead of -MD as
  dependencies on system headers are ignored by default
* When reviewing, make sure to remove stale .pyc files


obnoxxx:
Generally, the changes look good. They work for me (tm).

On the cosmetic side, the patch that touches third_party/waf should not be labeled
'build:wafsamba' in the commit message, but rather 'third_party:waf: ...'.

I don't know much about gcc deps, so I can't say much about the actual backported changes.
Since they are alreay upstream in waf 1.5, I tend to accept them.
Is anyone willing to look deeper?
The second and third patch are certainly OK.

abartlet:
I tried to look at these for review, but I'm also a bit lost in them.  I tend to accept ita's view with regard to build system patches. 

Can we get the cosmetics cleaned up and then I think obnoxxx and I can probably review them into Samba.

obnoxxx:
Looks good to me now.
Andrew? You ok, too?

obnoxxx:
withdrawing my RB for now:
configuring succeeds but build fails with backtrace and a message
"The gccdeps module has been upgraded. Please run distclean and reconfigure"

Can it be that the patch "use the upstream version.." needs to be changed as follows:


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 buildtools/wafsamba/wscript | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/buildtools/wafsamba/wscript b/buildtools/wafsamba/wscript
index 34775a6..03d16b1 100755
--- a/buildtools/wafsamba/wscript
+++ b/buildtools/wafsamba/wscript
@@ -554,7 +554,8 @@ def build(bld):
     if 'gccdeps' in sys.modules and (os.path.exists(p) or not bld.env.GCCDEPS):
         if os.path.exists(p):
             os.remove(p)
-        raise Utils.WafError('The gccdeps module has been upgraded. Please run distclean and reconfigure')
+        if bld.env.GCCDEPS:
+            raise Utils.WafError('The gccdeps module has been upgraded. Please run distclean and reconfigure')
 
     # give a more useful message if the source directory has moved
     relpath = os_path_relpath(bld.curdir, bld.srcnode.abspath())
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

ita1024:
The change you suggested is unnecessary:
1. The removal of `buildtools/wafsamba/gccdeps.py` went missing after the pull request was updated, so the old `gccdeps.py` file was still present, causing errors in the build.
2. For extra safety, the stale `gccdeps.pyc` is now removed before it is picked up in the configuration so that the first build will not raise any error.

obnoxxx:
Ok, cool. looks good. Review+

obnoxxx:
Andrew - still frightened? :-)

ita1024:
There is no hurry, more changes can be applied before this one and I am away this weekend. Thanks for applying all the others already.

GSam:
Simply attempting an autobuild with the patch fails during the initial build process. I did not have much time to investigate, @obnoxxx do you mind investigating?

ita1024:
No one wants to apply this, so we will submit the changes one by one instead (#43).

PR #33, opened 2015-11-05
build:wafsamba: Remove samba_utils.runonce

ita1024:
The decorator order matters in the following:
"""
@runonce
@conf
function()
"""

First, the @conf decorator binds the function to the configuration context
and then returns the same function. The @runonce decorator then takes
the output function and replaces it by another that performs caching. This new
function remains in the current script and is never bound to the configuration
context, so caching never occurs.

The declaration would have been correct if written like this:
"""
@conf
@runonce
function()
"""

Yet the decorator @run_once does not keep the function name (__name__), so the
annotation with @conf would fail. The straightforward approach is to remove
all the incorrect @runonce occurrences.

Besides that, the function Utils.run_once is already present in the Waf library,
and is sufficient for the current needs of the Samba code (it caches only the
first argument). Therefore samba_utils.runonce can be safely replaced by
Utils.run_once, which is also present in Waf 1.8.

Note: at runtime, only SETUP_BUILD_GROUPS seems to actually use caching.

### DIFF
--- a/buildtools/wafsamba/samba_autoconf.py
+++ b/buildtools/wafsamba/samba_autoconf.py
@@ -4,7 +4,7 @@ import os, sys
 import Build, Options, preproc, Logs
 from Configure import conf
 from TaskGen import feature
-from samba_utils import TO_LIST, GET_TARGET_TYPE, SET_TARGET_TYPE, runonce, unique_list, mkdir_p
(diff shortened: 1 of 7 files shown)


urisimchoni:
Reviewed-by: Uri Simchoni <uri@samba.org>

abartlet:
Merged as 963ccff806cf58898cdc8808138453d95df41b01

PR #42, opened 2015-12-04
Adds possibility to change another user password

GuillaumeGomez:
I sent the patch to `contributing@samba.org` already, but I'm not sure it's enough, so I post it here as well.

### DIFF
--- a/auth/credentials/credentials.c
+++ b/auth/credentials/credentials.c
@@ -43,6 +43,7 @@ _PUBLIC_ struct cli_credentials *cli_credentials_init(TALLOC_CTX *mem_ctx)
 
 	cred->workstation_obtained = CRED_UNINITIALISED;
 	cred->username_obtained = CRED_UNINITIALISED;
+	cred->user_to_connect_obtained = CRED_UNINITIALISED;
 	cred->password_obtained = CRED_UNINITIALISED;
 	cred->domain_obtained = CRED_UNINITIALISED;
 	cred->realm_obtained = CRED_UNINITIALISED;
@@ -51,15 +52,19 @@ _PUBLIC_ struct cli_credentials *cli_credentials_init(TALLOC_CTX *mem_ctx)
 	cred->principal_obtained = CRED_UNINITIALISED;
 	cred->keytab_obtained = CRED_UNINITIALISED;
 	cred->server_gss_creds_obtained = CRED_UNINITIALISED;
+	cred->user_to_connect_domain_obtained = CRED_UNINITIALISED;
 
 	cred->ccache_threshold = CRED_UNINITIALISED;
 	cred->client_gss_creds_threshold = CRED_UNINITIALISED;
 
 	cred->workstation = NULL;
 	cred->username = NULL;
+	cred->user_to_connect = NULL;
+	cred->password_to_connect = NULL;
 	cred->password = NULL;
 	cred->old_password = NULL;
 	cred->domain = NULL;
+	cred->user_to_connect_domain = NULL;
 	cred->realm = NULL;
 	cred->principal = NULL;
 	cred->salt_principal = NULL;
@@ -85,6 +90,8 @@ _PUBLIC_ struct cli_credentials *cli_credentials_init(TALLOC_CTX *mem_ctx)
 	cred->workstation_cb = NULL;
 	cred->password_cb = NULL;
 	cred->username_cb = NULL;
+	cred->user_to_connect_cb = NULL;
+	cred->user_to_connect_domain_cb = NULL;
 	cred->domain_cb = NULL;
 	cred->realm_cb = NULL;
 	cred->principal_cb = NULL;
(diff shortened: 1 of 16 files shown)


abartlet:
Thanks.  This (and samba-technical) is where we discuss patches, so you have indeed done the right thing.

In terms of the actual patch, can you investigate if the same kind of modification could be made to samba-tool or net instead? smbpasswd is essentially frozen in time at this point, new functionality is added to our more general (and esaiser to expand) tools.

In particular, this is because smbpasswd has enough of a non-standard command-line syntax as it is (inconsistent with the rest of samba, and getopt rather than popt based).

Sorry,

Andrew Bartlett

GuillaumeGomez:
The change is only on smbpasswd and is what I need. I didn't see any call for the feature I modified for out library in smbpasswd, that's why I changed it this way

GuillaumeGomez:
So what will happen now ? I only updated smbpassd source code because other used libs' functions already fit what I needed. I don't see how I could do it differently...

abartlet:
I understand it was the easiest to change, but we sadly can't accept such a patch to smbpasswd.

The easiest patch to write is sadly often not the best patch for the project as a whole.

I would suggest finding a way to patch 'samba-tool user password' to cover the auth/change user split you are after.  It is a combination of python and C, but perhaps you can patch py_net_change_password in python/samba/netcmd/py_net.c and python/samba/netcmd/user.py to have such an option.

GuillaumeGomez:
I don't get it. Why should I update samba-tool when I'm using smbpasswd ? Is samba-tool going to replace smbpasswd ?

GuillaumeGomez:
Ok, as I understood it, samba-tool is the replacement of smbpasswd (or at least, does what smbpasswd did), so I'll update my patch in consequence.

GuillaumeGomez:
(Happy new year !). I added the change on smbpasswd, so now, each version is in its own commit.

GuillaumeGomez:
Any news on this pull request?

aaptel:
Additionally, since the logon name has a maximum length your patch could be simplified.

From https://technet.microsoft.com/it-it/library/bb726984%28en-us%29.aspx:
>Logon names can be up to 104 characters. However, it isn't practical to use logon names that are longer than 64 characters.

GuillaumeGomez:
I wanted to make the change as little as possible but that seems to be a good idea. Thanks @aaptel!

abartlet:
On Mon, 2016-01-18 at 02:20 -0800, Guillaume Gomez wrote:
> Any news on this pull request?
> —
> Reply to this email directly or view it on GitHub.

This pull request isn't going anywhere, sorry.

I know it isn't simple to patch our other tools, but smbpasswd is *not*
going to get another option, even one as 'simple' as this.  It is
legacy tool kept because it is often used and may be referenced in
scripts.  The new tools are 'net' and 'samba-tool'.

Sorry,

Andrew Bartlett

-- 
Andrew Bartlett                       http://samba.org/~abartlet/
Authentication Developer, Samba Team  http://samba.org
Samba Developer, Catalyst IT          http://catalyst.net.nz/services/samba

GuillaumeGomez:
@abartlet: That's what I did in the second commit, I updated 'samba-tool'. If you want, I can remove the first one from the logs to allow you to take a simpler look.

hortimech:
Can somebody explain why the code in source4 needs to be altered when samba-tool already does what GuillaumeGomez is trying to do

i.e.
samba-tool user setpassword --help
Usage: samba-tool user setpassword (<username>|--filter <filter>) [options]

Set or reset the password of a user account.

The command may be run from the root userid or another authorized userid.  The
-H or --URL= option can be used to execute the command against a remote
server.

I am not adverse to changes if they are required and helpful, but are they needed here ?

GuillaumeGomez:
I kept the first one to allow to easily understand what the second does (the version 4 of samba source code is very difficult to read and understand). Anyway, I removed the first commit, it only remains the source 4 commit.

GuillaumeGomez:
@hortimech: Hum... Is it really the same thing? I took a look and didn't find any "obvious" way to do the thing I was looking for (and as far as I know, it wasn't possible in version 3, and that's the original reason of this PR).

hortimech:
As far as I am aware, the code in source4 is for Samba 4.x.x and samba-tool only works with Samba 4.x.x. I think you are also missing the fact that Samba 3.x.x is EOL.

I have no idea if your code is good code or otherwise, I just don't want something changing for a corner case if this may adversely affect others.

GuillaumeGomez:
Well, it's your call. Just keep me up to date about your final decision please.

bjacke:
you are right, that samba-tool is not currently the tool useful on Samba servers in "classic" mode, so your request for enhancing with a patch was not wrong. But I think the functionality is there already: as root you can just run "smbpasswd foo" and as a normal user you can change someone else's password with "smbpasswd -U foo".

https://bugzilla.samba.org/ is the tool you shuld use to report bugs and feature requests, please use that next time.

Thanks
Björn

bjacke:
you're right, that samba-tool is not yet very useful for Samba servers in "classic" mode and your request to have that feature in smbpasswd is quite valid. But the feature is already implementen as far as I can see: as root you can run "smbpasswd foo" to set someone's password and as a user you can set someone else's password with "smbpasswd -U foo", can't you?

https://bugzilla.samba.org/ is the right way to report bugs and feature requests for Samba, please use that in the future.

Thanks
Björn

PR #44, opened 2015-12-10
Add option for overlayfs "disable openfile inode check"

RG72:
Add a option disable openfile inode check.

I use samba over overlayfs. 
When users try to open file for write, they will get NT_STATUS_ACCESS_DENIED and file opens for read only. 
Because open_file_ntcreate: file %s - dev/ino mismatch.

With disable openfile inode check = true, users can openfile for write.
I build a time machine of share, with overlayfs and hard links.


douglasbagnall:
This proposal was fairly thoroughly rejected on the mailing list:
https://lists.samba.org/archive/samba-technical/2015-December/111125.html

Can someone with closing powers close this pull request?

Sorry @RG72 if our technical communication failures left you wondering for three years.

PR #45, opened 2015-12-10
Added MSV1_0_ALLOW_MSVCHAPV2 flag to ntlm_auth

qnet-herwin:
An implementation of https://lists.samba.org/archive/samba/2012-March/166497.html (which has been discussed in 2012, but was never implemented).

It has been tested on a Debian Jessie system with this patch added to the Debian package (which is currently 4.1.17). Even though this is Samba 4, the ntlm_auth installed is the one from Samba 3 (yes, it surprised me too). The backend was a machine with Windows 2012R2.

It was first tested with the local security policy 'Network Security: LAN Manager authentication level' setting changed to 'Send NTLMv2 Response Only' (allow ntlm v1). This way we are able to authenticate with and without the MSV1_0_ALLOW_MSVCHAPV2 flag (as expected).

After the basic step has been verified, the local security policy 'Network Security: LAN Manager authentication level' setting was changed to 'Send NTLMv2 Response Only. Refuse LM & NTLM' (only allow ntlm v2). The behaviour now changed according to the MSV1_0_ALLOW_MSVCHAPV2 flag (again: as expected).

```
$ ntlm_auth --request-nt-key --username=XXXXXXXXXXXXX --challenge=XXXXXXXXXXXXXXXXX --nt-response=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX --domain=
Logon failure (0xc000006d)
$ ntlm_auth --request-nt-key --username=XXXXXXXXXXXXX --challenge=XXXXXXXXXXXXXXXXX --nt-response=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX --domain= --allow-mschapv2
NT_KEY: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

The changes in `wbclient.h` are intended for programs that use libwinbind directly instead of authenticating via `ntlm_auth`. I intend to use that within FreeRADIUS (see https://bugzilla.samba.org/show_bug.cgi?id=11149).


abartlet:
Thanks so much for following up on what I dropped the ball on, and for testing it with actual windows (Samba will still ban such a login as 'not NTLMv2' per the smb.conf if it is the DC).

qnet-herwin:
I've added a comparable patch to FreeRADIUS (see https://github.com/FreeRADIUS/freeradius-server/commit/37f2f6d8e09bdebdf3031e419c00a0d3193b074a), where it arose the question if this flag shouldn't be enabled as default. The original commit on the mailing lists added is as default instead of an option, and the flags `MSV1_0_ALLOW_WORKSTATION_TRUST_ACCOUNT` and `MSV1_0_ALLOW_SERVER_TRUST_ACCOUNT` are hardcoded in the source as well.

It is a behavioural change, and I honestly have no idea if this could break anything. Hardcoding the flag would definitely result in cleaner code.

abartlet:
It has to be optional, as we don't want an NTLM login on some other protocol somehow being routed via ntlm_auth triggering it. The whole thing is a farce, but one we have to live with.

abartlet:
To be clear, I mean that it should be optional in ntlm_auth.  I'll comment on the FreeRADIUS side there.

mcnewton:
This looks great to me, @qnet-herwin - thanks for doing something that's been on my TODO list since March :-). After spending the first three months of 2015 adding the Ctx calls to libwbclient and the FreeRADIUS code, I totally ran out of steam on this. I also figured I should catch up on real work... ;-)

In my opinion, make it optional for ntlm_auth, but hardcoded in FreeRADIUS.

qnet-herwin:
I don't want to appear impatient, but is there anything else that needs to happen before this change can be merged?

abartlet:
Can you add your Signed-off-by: to the commit message?

Ideally also add such a mode to the ntlm-server-1 mode.  I had hoped FreeRADIUS would have moved to that mode by now, it was added for FreeRADIUS as better option than command-line parameters :-)

https://www.samba.org/samba/devel/copyright-policy.html
https://wiki.samba.org/index.php/CodeReview

Sorry for dropping the ball on this.

I would love a test (which won't make much difference as our server-side doesn't respect it, but will at least ensure the option keeps working), which will mean calculating the challenge/response values matching the selftest passwords from selftest/target/.

I realise that is a bigger ask.

Andrew Bartlett

abartlet:
If we at least get the signed-off-by, I can get this much in before the 4.4 branch closes on 26 Jan.

mcnewton:
@abartlet - I updated FreeRADIUS last year to use libwbclient directly so no need to communicate ntlm_auth at all now (see https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_mschap/auth_wbclient.c#L123). So @qnet-herwin has added the WBC_MSV1_0_ALLOW_MSVCHAPV2 flag to the wbcCtxAuthenticateUserEx call there. It seems to work much faster than execing ntlm_auth each time. :-)

Could just do with an option to disable all of winbind's UID mapping/caching to remove another overhead, but haven't had time to look at this.

Aside from historical reasons and people using old software, the only reason for FreeRADIUS to use ntlm_auth now is for password changes (which uses the ntlm-change-password-1 mode), but this could also be able to be changed to use libwbclient as well at some point.

Cheers,
Matthew

qnet-herwin:
I've just pushed it again with the Signed-Off-thingy. I'll try to have a look at the tests, but can't promise anything.

qnet-herwin:
Regarding the coupling between FreeRADIUS and Samba (as @mcnewton mentioned): Yes, it is preferable to use libwbclient directly, but I would still very much like to have WBC_MSV1_0_ALLOW_MSVCHAPV2 exported and get rid of these lines: https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_mschap/auth_wbclient.h#L10-L13

abartlet:

It would be great for this to be in 4.4, can I get a second team reviewer please?

Thanks,

Andrew Bartlett

kblin:
Personally I'd prefer if the if statements had braces, even if they're just containing a single line right now.

qnet-herwin:
@kblin: The change at https://github.com/samba-team/samba/pull/45/files#diff-9cc1292d5d9c92f82114e9221ff610e0R531 is at least in the same style as the line above (which also misses braces). The change at https://github.com/samba-team/samba/pull/45/files#diff-5c75b7b216d95dfdf084e1fa404ec7baR188 doesn't have any single-line if-statements close to it (but the style was modelled after the samba3-version).

I think adding braces to samba3/ntlm_auth should be done in a separate commit

kblin:
But I guess this doesn't invalidate the code technically. So add my reviewed by, please.

kblin:
Yes, I know. But just because the rest of the ntlm_auth code sucks, doesn't mean new code needs to suck ;). But seeing how the deadline to get this into 4.4 is today, let's fix that later.

nicklowe:
Any reason this is still open?

qnet-herwin:
Looks as if this has been merged in commit 0b500d413c5b76188c0c566318be7079b777237c, so I'll close this case.

PR #46, opened 2015-12-15
Remove posixAccount from samdb.py

hortimech:
This will stop the adding of the totally unnecessary  posixAccount objectClass when adding/updating a user.

Signed-off-by: Rowland Penny rpenny@samba.org


abartlet:
Please just have one patch topic per pull request, the SOA records are a different, more complex matter.

GSam:
Merged as c16e770d9670794abc8dc08b62e2c01197bd97c5

PR #47, opened 2015-12-18
If samlogon for trusted child domain user fails attempt to reroute re…

noelpower:
…quest

When kerboros authentication fails we may attempt to fallback to
samlogon.  However schannel netlogon connections from a domain child
winbindd to the domain controller when that domain is not 'our'
domain are dissallowed and thus the credentials are not available.
The samlogon request when this happens cannot be serviced. This patch
detects if the samlogon fallback will occur for a non primary domain
winbindd child, in this case it will return a status of
NT_STATUS_MORE_PROCESSING_REQUIRED to the parent.
The parent then will then retry the authentication by chosing and sending
the request to a domain child that should be able to handle it.

Signed-off-by: Noel Power noel.power@suse.com


abartlet:
I'm confused why we don't route the request correctly in the first place?  Is it because we hope for Kerberos?

Either way, please reopen on GitLab. 

Thanks!

PR #48, opened 2015-12-28
Make several libraries private.

jelmer:
This makes several libraries that are currently installed by Samba private, and reduces the number of header files that is installed.

None of these libraries have a stable API/ABI, and I am not aware of any external users of them.


jelmer:
Updated to take into account metze's comments.

abartlet:
This looks good to me.

PR #49, opened 2016-01-04
Cloud build

abartlet:
I would like to have Samba start being tested on travis-ci.org, so that we know if pull requests break the tree before they are reviewed, not after.

I realise that the make -j in the static and samba-libs builds may be an issue, but otherwise these come far to close to the 50min timeout (they take 23mins with -j). I have already split up the samba-libs build into samba-libs and samba-static for similar reasons.

The main samba build and test is not done at this time, as we need to come up with a new quicktest that can run in 20mins or so.

The autobuild.py script has been adjusted to use cp --link rather than git checkout to reduce disk use.

The CTDB and samba builds have been omitted until they are reliable.

Please review/push

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ env:
   - TASK=tevent
   - TASK=pidl
 
+# Fail everything after the first job fails
+matrix:
+  fast_finish: true
+
 before_install:
  - sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


abartlet:
Merged as ff94a01e19604ae6275672e70edd25121df79bbf

PR #50, opened 2016-01-06
Python ldb Py_ssize_t fixes

abartlet:
This should fix issues on s390x in recent ldb, by being strict about always using py_ssize_t

I also clean up some code supporting Python 2.4, which we no longer support


jelmer:
Reviewed-By: me

Thanks!

abartlet:
Merged as 78e9f1effb0d8f5fd4eb170ff44af428bc5c2315

PR #51, opened 2016-01-12
fixed cpp warning due to missing #ifdef directive in lib/util/debug.h

krichter722:


abartlet:
The travis-ci failure: https://travis-ci.org/samba-team/samba/jobs/101922474 appears to be be a red herring.

abartlet:
My apologies for the very long delay in dealing with this. 

I really don't think this patch is right.  If DEBUG is already defined, then why not all the others.  My guess is an unrelated DEBUG with a different meaning is defined in this build environment. 

Either way, our web-based development tooling has moved to GitLab, so we are clearing out the old Pull Requests here. 

Sorry,

Andrew Bartlett

PR #52, opened 2016-01-14
Stop installing more headers without well-maintained API stability.

jelmer:


b-a-t:
Is this update still valid?

jelmer:
No, this is stale - I'll remove it.

PR #53, opened 2016-02-12
waf: Prefer libsystemd to deprecated libsystemd-*

heftig:
Not actually tested on a system that only has the deprecated libs.


cryptomilk:
This has been addressed with 09bc5b5374227a555f580c3d7c1d82f15bb818bc

douglasbagnall:
Already fixed, as per @cryptomilk's comment.

PR #54, opened 2016-02-24
acess based share enum: handle permission set in configuration files

bud4:
** access based share enum** not work with permission set in config file.
change function is_enumeration_allowed to check  permissions set by
fields: valid users, invalid users, only user.

Signed-off-by: Alberto Maria Fiaschi <alberto.fiaschi@estar.toscana.it>

### DIFF
--- a/source3/rpc_server/srvsvc/srv_srvsvc_nt.c
+++ b/source3/rpc_server/srvsvc/srv_srvsvc_nt.c
@@ -480,8 +480,13 @@ static bool is_enumeration_allowed(struct pipes_struct *p,
     if (!lp_access_based_share_enum(snum))
         return true;
 
+    if (!user_ok_token(p->session_info->unix_info->unix_name,
+                       p->session_info->info->domain_name,
+                       p->session_info->security_token, snum))
+        return false;
+
     return share_access_check(p->session_info->security_token,
-			      lp_servicename(talloc_tos(), snum),
+                              lp_servicename(talloc_tos(), snum),
 			      FILE_READ_DATA, NULL);
 }
 


urisimchoni:
Hi Alberto,
Thanks for submitting this fix.
Some comments:

1. The checks for lp_valid_users and lp_invalid_users seem redundant - they are being done inside user_ok_token() without extra cost, it also seems that the proposed patch misses lp_only_user() this way.

2. The "username" parameter to user_ok_token() needs to be session_info->unix_info->unix_name (that's what you exchange for %U)

3. We need a bug in bugzilla and a BUG: line pointing to it in the commit message (see other commit messages in git log)

4. Please consult README.Coding - surround if blocks with curly braces. The preferred coding style also prefers multiple simple if statements over a complex condition, if possible. Something like:
```c
if (!user_ok_token(...)) {
        return false;
}

return share_access_check(...);
```

Could you resubmit with those changes? I can later throw in some tests to verify correct behavior.
Thanks,
Uri

bud4:
2016-02-28 13:28 GMT+01:00 Uri Simchoni <notifications@github.com>:

> Hi Alberto,
> Thanks for submitting this fix.
> Some comments:
>
>    1.
>
>    The checks for lp_valid_users and lp_invalid_users seem redundant -
>    they are being done inside user_ok_token() without extra cost, it also
>    seems that the proposed patch misses lp_only_user() this way.
>
>     sorry I did not see that the user_ok_token function returns true if in
the configuration file is not set any permission.


>
>    1.
>
>    The "username" parameter to user_ok_token() needs to be
>    session_info->unix_info->unix_name (that's what you exchange for %U)
>
> ok

>
>    1.
>
>    We need a bug in bugzilla and a BUG: line pointing to it in the commit
>    message (see other commit messages in git log)
>
> ok if the bug is not present in bugzilla then I'll create it

>
>    1.
>
>    Please consult README.Coding - surround if blocks with curly braces.
>    The preferred coding style also prefers multiple simple if statements over
>    a complex condition, if possible. Something like:
>
> if (!user_ok_token(...)) {
>         return false;
> }
> return share_access_check(...);
>
> ok i will do it as soon as I can...
Bye
Alberto

Could you resubmit with those changes? I can later throw in some tests to

urisimchoni:
I replied in samba-technical list with patch that's essentially the same, save for coding style fixes, plus a blackbox test. Awaiting Alberto's approval and another reviewer.

bud4:
for me ok
Alberto

2016-03-01 22:44 GMT+01:00 Uri Simchoni <notifications@github.com>:

urisimchoni:
Merged as 5036a0922b7890005bcc8b77368a6635c8ebeb4b (with some indentation / code style modifications)

PR #55, opened 2016-02-28
ctdb: Fix detection of gnukfreebsd

stevenc99:
GNU/kFreeBSD's platform name is 'gnukfreebsd', not just 'kfreebsd'.

I originally sent this patch to https://bugs.debian.org/802621#15
where I was asked to send it upstream.  Thanks!

### DIFF
--- a/ctdb/wscript
+++ b/ctdb/wscript
@@ -315,7 +315,7 @@ def build(bld):
         CTDB_SYSTEM_SRC = bld.SUBDIR('common', 'system_aix.c')
     elif sys.platform.startswith('freebsd'):
         CTDB_SYSTEM_SRC = bld.SUBDIR('common', 'system_freebsd.c')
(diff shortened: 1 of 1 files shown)


GSam:
Merged as d692020c1aa75ca1f1950d451937334766da238f

sathieu:
Can this patch be applied to previous branches?

stevenc99:
Hello,

Yes the bug is present since v4-2-stable and the patch can be applied to
all stable branches since that one.

Thanks,
Regards,
-- 
Steven Chamberlain
steven@pyro.eu.org

sathieu:
OK. I'll apply it directly to the Debian package then.

Cheers

martin-schwenke:
If you open that bug then we'll backport it and save you the trouble...  :-)

PR #56, opened 2016-03-26
build:wafsamba: Remove ambiguous 'if x in conf.env' constructs

ita1024:
Configuration values such as HAVE_STDDEF_H can be set to 0
to indicate a test failure. Waf 1.5 has a few bugs that
prevent configuration tests from setting such values
consistently on failures.

Consequently, conditions such as 'if conf.env.VARNAME' must be
used to indicate that config test successes are expected.
Note that conf.env.VARNAME always returns an empty list (False value)
when no variable is defined so there are no risk of raising
AttributeError/KeyError exceptions.


urisimchoni:
Merged as e36e1f7b12291ac818206cadc2b3081a02f62f84

PR #57, opened 2016-04-01
Fix bug #11658 - ZFS snapshots access denied

jackqq:
Check the return status of smbd_check_access_rights more precisely, to allow access to ZFS snapshots, where NT_STATUS_NOT_SUPPORTED is returned and previously regarded as access denied.

Signed-off-by: QIU Quan <jackqq@gmail.com>


jackqq:
Renamed fix branch name and requested as #58.

PR #59, opened 2016-04-09
Correctly set cli->raw_status for libsmbclient in SMB2 code

Xenopathic:
The SMB2 file handling code wasn't correctly setting raw_status, which is used by libsmbclient to report file open errors etc.

The bug was originally found when attempting to do a `stat()` on a non-existing file behind a DFS link. Before this patch, the code returned was 22 (EINVAL), rather than the expected 2 (ENOENT), since the SMB result status wasn't set into the cli object. The easiest way to demonstrate this is to run the attached test program against a Samba server configured with DFS (for an easy test environment, run `docker pull xenopathic/samba-dfs && docker run -e SMB_USER=test -e SMB_PASSWORD=test --rm xenopathic/samba-dfs`). Make sure smb.conf contains `client max protocol = SMB2` or higher of course.

The fix for the actual noticed bug is the line added to `cli_smb2_create_fnum_recv()`, but I added similar lines to all other `cli_smb2_*_recv()` functions since the behaviour should be the same.

Test program:

```
#include <libsmbclient.h>
#include <stdio.h>
#include <errno.h>
#include <string.h>

const char *guser = "test";
const char *gpass = "test";
const char *url = "smb://172.17.0.2/dfs/public/Test.pdf";

void authfn(SMBCCTX *ctx, const char *server, const char *share, char *wkgp, int wkgplen, char *user, int userlen, char *pass, int passlen) {
    strcpy(user, guser);
    strcpy(pass, gpass);
}

int main() {
    SMBCCTX *ctx = smbc_new_context();
    smbc_setFunctionAuthDataWithContext(ctx, authfn);
    ctx = smbc_init_context(ctx);

    smbc_stat_fn smbc_stat = smbc_getFunctionStat(ctx);

    struct stat statbuf;
    printf("smbc_stat(): %d\n", smbc_stat(ctx, url, &statbuf));
    printf("errno: %d\n", errno);

    smbc_free_context(ctx, 1);

    return 0;
}
```

Signed-off-by: Robin McCorkell robin@mccorkell.me.uk

### DIFF
--- a/source3/libsmb/cli_smb2_fnum.c
+++ b/source3/libsmb/cli_smb2_fnum.c
@@ -262,6 +262,7 @@ NTSTATUS cli_smb2_create_fnum_recv(struct tevent_req *req, uint16_t *pfnum,
 	NTSTATUS status;
 
 	if (tevent_req_is_nterror(req, &status)) {
+		state->cli->raw_status = status;
 		return status;
 	}
 	if (pfnum != NULL) {
@@ -270,6 +271,7 @@ NTSTATUS cli_smb2_create_fnum_recv(struct tevent_req *req, uint16_t *pfnum,
 	if (cr != NULL) {
 		*cr = state->cr;
 	}
+	state->cli->raw_status = NT_STATUS_OK;
 	return NT_STATUS_OK;
 }
 
@@ -390,7 +392,11 @@ static void cli_smb2_close_fnum_done(struct tevent_req *subreq)
 
 NTSTATUS cli_smb2_close_fnum_recv(struct tevent_req *req)
 {
-	return tevent_req_simple_recv_ntstatus(req);
+	struct cli_smb2_close_fnum_state *state = tevent_req_data(
+		req, struct cli_smb2_close_fnum_state);
+	NTSTATUS status = tevent_req_simple_recv_ntstatus(req);
+	state->cli->raw_status = status;
+	return status;
 }
 
 NTSTATUS cli_smb2_close_fnum(struct cli_state *cli, uint16_t fnum)
@@ -2302,6 +2308,7 @@ NTSTATUS cli_smb2_read_recv(struct tevent_req *req,
 				req, struct cli_smb2_read_state);
 
 	if (tevent_req_is_nterror(req, &status)) {
+		state->cli->raw_status = status;
 		return status;
 	}
 	/*
@@ -2311,6 +2318,7 @@ NTSTATUS cli_smb2_read_recv(struct tevent_req *req,
 	 */
 	*received = (ssize_t)state->received;
 	*rcvbuf = state->buf;
+	state->cli->raw_status = NT_STATUS_OK;
 	return NT_STATUS_OK;
 }
 
@@ -2409,6 +2417,7 @@ NTSTATUS cli_smb2_write_recv(struct tevent_req *req,
 	NTSTATUS status;
 
 	if (tevent_req_is_nterror(req, &status)) {
+		state->cli->raw_status = status;
 		tevent_req_received(req);
 		return status;
 	}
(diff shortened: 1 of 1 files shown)


jrasamba:
On Sat, Apr 09, 2016 at 05:37:31AM -0700, Robin McCorkell wrote:
> The SMB2 file handling code wasn't correctly setting raw_status, which is used by libsmbclient to report file open errors etc.
> 
> The bug was originally found when attempting to do a `stat()` on a non-existing file behind a DFS link. Before this patch, the code returned was 22 (EINVAL), rather than the expected 2 (ENOENT), since the SMB result status wasn't set into the cli object. The easiest way to demonstrate this is to run the attached test program against a Samba server configured with DFS (for an easy test environment, run `docker pull xenopathic/samba-dfs && docker run -e SMB_USER=test -e SMB_PASSWORD=test --rm xenopathic/samba-dfs`). Make sure smb.conf contains `client max protocol = SMB2` or higher of course.
> 
> The fix for the actual noticed bug is the line added to `cli_smb2_create_fnum_recv()`, but I added similar lines to all other `cli_smb2_*_recv()` functions since the behaviour should be the same.

This is close, but not quite in the right place I think.

Can you give me a little time to look into this ?

Did you log a bug in bugzilla.samba.org so we can track it?

Cheers,

	Jeremy.

Xenopathic:
No, I didn't post a bug report, since I've contributed code in the past so just directly debugged the problem. Let me do that now.

From analysis of other functions that set raw_status, I thought that _recv() was the best place to put the new code. Alternatively, _done() might be a better place?

jrasamba:
On Wed, Apr 13, 2016 at 02:11:30PM -0700, Robin McCorkell wrote:
> No, I didn't post a bug report, since I've contributed code in the past so just directly debugged the problem. Let me do that now.

Thanks.

> From analysis of other functions that set raw_status, I thought that _recv() was the best place to put the new code. Alternatively, _done() might be a better place?

Yeah, I'm thinking about cleaning up all the
places that set raw_status to centralize things
so we fix it everywhere.

There are probably some SMB1 calls that still
don't do this right...

Xenopathic:
Bug reported at https://bugzilla.samba.org/show_bug.cgi?id=11834

Xenopathic:
Any update? Point me in the right direction and I'll create a patch to implement the raw_status stuff as it should be. At the moment though, the SMB2 behaviour is broken...

jrasamba:
On Wed, Apr 20, 2016 at 04:05:08AM -0700, Robin McCorkell wrote:
> Any update? Point me in the right direction and I'll create a patch to implement the raw_status stuff as it should be. At the moment though, the SMB2 behaviour is broken...

Sorry, out at the Vault conf. this week
with poor access to email (and busy too :-).

Ping me again next week if I haven't gotten
back to you by then.

jrasamba:
On Wed, Apr 20, 2016 at 04:05:08AM -0700, Robin McCorkell wrote:
> Any update? Point me in the right direction and I'll create a patch to implement the raw_status stuff as it should be. At the moment though, the SMB2 behaviour is broken...

OK, raw_status should be set in the async _recv() functions
inside source3/libsmb/cli_smb2_fnum.c I think.

Xenopathic:
So this patch is good then?

jrasamba:
On Sat, Apr 30, 2016 at 04:23:58AM -0700, Robin McCorkell wrote:
> So this patch is good then?

ENOPATCHATTACHED :-).

Xenopathic:
I submitted the patch through GitHub, so it's the one attached to the first mailing list email

jrasamba:
On Mon, May 02, 2016 at 11:49:20AM -0700, Robin McCorkell wrote:
> I submitted the patch through GitHub, so it's the one attached to the first mailing list email

Send directly to the samba-technical mailing list. I'm less likely
to miss it that way.

Xenopathic:
Ooo, this is an old PR. Already merged via samba-technical -> closing

PR #60, opened 2016-04-26
Add build option for default smbpasswd location

sathieu:
This part of the FHS patch was never submitted from https://bugs.debian.org/705449

Signed-off-by: Ivo De Decker ivo.dedecker@ugent.be


sathieu:
Done in 685e813c45436046a18eb3b8895da5846bd2a0dc (samba-4.5.0rc1).

PR #65, opened 2016-10-24
Samba tests credentials

frenzymadness:
Hello.

My name is Lumír Balhar and I am from Python maintenance team in Red Hat. I'm taking over Petr Viktorin's effort to port Samba to Python 3, since he's quite busy right now.

This is pull request improves test coverage of the credentials module, and fixes DeprecationWarning in the same test case. Everything only for Python 2.7.

I am looking forward to cooperation.
Lumír

### DIFF
--- a/python/samba/tests/credentials.py
+++ b/python/samba/tests/credentials.py
@@ -32,35 +32,35 @@ class CredentialsTests(samba.tests.TestCase):
 
     def test_set_username(self):
         self.creds.set_username("somebody")
-        self.assertEquals("somebody", self.creds.get_username())
+        self.assertEqual("somebody", self.creds.get_username())
 
     def test_set_password(self):
         self.creds.set_password("S3CreT")
-        self.assertEquals("S3CreT", self.creds.get_password())
+        self.assertEqual("S3CreT", self.creds.get_password())
 
     def test_set_domain(self):
         self.creds.set_domain("ABMAS")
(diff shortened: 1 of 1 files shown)


frenzymadness:
Hello.

Thank you for review.

The reason for usage of 'utf-8-sig' is that result of functions `get_password()` and `get_old_password()` starts with BOM Unicode character [0] in this case.

[0] https://en.wikipedia.org/wiki/Byte_order_mark

Have a nice day.
Lumir

frenzymadness:
Hello.

I think that it isn't a bug because BOM is in input data for *set* functions. BOM is automatically added when you decode string to UTF-16.

For example:
```
>>> pwd = 'TesT'
>>> pwd.encode('utf-16')
'\xff\xfeT\x00e\x00s\x00T\x00'
```

So I think that in this case is the output of *get* functions correct.

What do you think about it?

frenzymadness:
Hello, Andrew.

I fixed badly chosen encoding and now everything looks better and works.

Do you think that my PR is ok now?

Thank you very much for your review and have a nice day.
Lumír

frenzymadness:
Hello.

Is there any problem with merge?

Have a nice day.
Lumír

GSam:
Merged in commit a00d34dbc1cf0348be2417ee21903ab1bbf77a3d

PR #67, opened 2016-10-27
vfs_gluster: glfs_clear_preopened: fix leaked elements

pkalever:
It is glfs_clear_preopened() responsibility to clean memory allocated in
glfs_set_preopened()

Signed-off-by: Prasanna Kumar Kalever prasanna.kalever@redhat.com

### DIFF
--- a/source3/modules/vfs_glusterfs.c
+++ b/source3/modules/vfs_glusterfs.c
@@ -152,6 +152,9 @@ static void glfs_clear_preopened(glfs_t *fs)
 			DLIST_REMOVE(glfs_preopened, entry);
 
 			glfs_fini(entry->fs);
+			
+			talloc_free(entry->connectpath);
+			talloc_free(entry->volume);
 			talloc_free(entry);
 		}
 	}


source3/modules/vfs_glusterfs.c
```diff
@@ -152,6 +152,9 @@ static void glfs_clear_preopened(glfs_t *fs)
 			DLIST_REMOVE(glfs_preopened, entry);
 
 			glfs_fini(entry->fs);
+			
+			talloc_free(entry->connectpath);
+			talloc_free(entry->volume);
 			talloc_free(entry);
```
obnoxxx: The nice thing about talloc is the hierarchy:  since the connectpath and volume members are allocated on the parent context "entry", it is sufficient to call `talloc_free(entry);`

obnoxxx:
@pkalever, why is there a merge commit in the PR? This is strange...

BTW: the preferred way of doing samba contributions is sending git-formatted patchsets by mail to samba-technical@lists.samba.org .

source3/modules/vfs_glusterfs.c
pkalever: Wow that's cool!
Thanks for teaching this @obnoxxx

PR #69, opened 2016-11-23
Samba bug 12443: Active Directory account locked when using winbind refresh tickets

dmulder:
Resolves an issue where user accounts get locked out due to winbind refreshing tickets using cached passwords (after the password has been modified, but the wrong password is still cached).

It's my opinion that the password kinit should be disabled by default. Does anyone disagree?


dmulder:
I suppose this change begs the question, should this situation also be handled in krb5_ticket_gain_handler()?

dmulder:
Sorry, I'm using this as WIP branch

PR #70, opened 2016-11-30
packaging: Avoid timeout for nmbd if started offline with systemd

mtdcr:
If no network connection appears within DefaultTimeoutStartSec
(~ 90s) after startup, nmbd fails to notify systemd and will
therefore get killed.

Signed-off-by: Andreas Oberritter <obi@opendreambox.org>

### DIFF
--- a/packaging/systemd/nmb.service
+++ b/packaging/systemd/nmb.service
@@ -10,6 +10,7 @@ EnvironmentFile=-/etc/sysconfig/samba
 ExecStart=/usr/sbin/nmbd $NMBDOPTIONS
 ExecReload=/usr/bin/kill -HUP $MAINPID
 LimitCORE=infinity
+TimeoutStartSec=0
 
 [Install]
 WantedBy=multi-user.target


mtdcr:
Given the simplicity of this patch, I wonder how long I should wait before annoying _those with write access_ with repeated queries about the status of this pull request.

abartlet:
I'm very sorry for the delay.

I do wonder however if it is as simple as you suggest - shouldn't we have Samba indicate that it has started (pending any interfaces) rather than disabling the whole startup signalling mechanism?

mtdcr:
On Mon, 20 Feb 2017 20:04:56 -0800
Andrew Bartlett <notifications@github.com> wrote:

> I'm very sorry for the delay.
> 
> I do wonder however if it is as simple as you suggest - shouldn't we have Samba indicate that it has started (pending any interfaces) rather than disabling the whole startup signalling mechanism?
> 

The signalling works fine and therefore smbd doesn't get started before nmbd finds a network connection.

cryptomilk:
This is the wrong approach:

See https://bugzilla.samba.org/show_bug.cgi?id=13184

abartlet:
Closing in favour of 0e571054a61e9de69190ae023199d1670e097e88 in master for Samba 4.8

PR #71, opened 2017-01-19
Add missing import of sys in ms_schema.py

andiges:
If we get an error, we try to print it, but sys module is missing


urisimchoni:
Merged as f73c6bd2698ccd6e52b0782407c1c6c6fe78d60d (with a slight whitespace fix)

PR #75, opened 2017-01-30
Partially roll back f9e1887477 so that 'clean' completes (CI test only - do not merge)

axs-gentoo:


axs-gentoo:
Giving up, this is above my ability to fix with fumbling around...

PR #76, opened 2017-01-30
    Fix FreeBSD/Darwin & Linux sendfile.

j-c-m:
Use poll() to emulate a blocking sendfile().

    This avoids spinning on EAGAIN/EWOULDBLOCK under Linux implementation,
    which greatly reduces CPU usage.

(https://lists.samba.org/archive/samba/2013-September/175838.html)

    This fixes FreeBSD/Darwin implementation, which under most cases did
    not work, it would break early when nwritten == 0.  Which would happen
    anytime it spun on EAGAIN writing 0 bytes.

(https://forums.freebsd.org/threads/55017/)

### DIFF
--- a/source3/lib/sendfile.c
+++ b/source3/lib/sendfile.c
@@ -106,7 +106,7 @@ ssize_t sys_sendfile_native(int tofd, int fromfd, off_t offset, size_t count, st
 
 	if (hv->iov_len > 0) {
 		if((nwritten = send(tofd, hv->iov_base, hv->iov_len, MSG_MORE)) == -1) {
-			if(errno != EINTR || errno != EAGAIN || errno != EWOULDBLOCK) {
+			if(errno != EINTR && errno != EAGAIN && errno != EWOULDBLOCK) {
 				return -1;
 			} else {
 				return 0;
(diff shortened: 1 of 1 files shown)


bjacke:
please use max 80 chars/line, see README.Coding. And there is more obvious wrong at two places:
"if(errno != EINTR || errno != EAGAIN || errno != EWOULDBLOCK)" is always true. Can you fix those things and try to explain your code changes so that reviewing is better possible? Finally, please send the patch to the samba-technical mailing list because github is not working well with the samba community for communication and discussion.
Thank you - Björn

bjacke:
should be fixed with 456e520a3be7e4b54f1f144324c3671b8f6e35ea - can you cross-check that the latests 4.8 or 4.9 releases work fine for you now?

j-c-m:
Yeah it looks like that would fix the issue, I haven't tested.

bjacke:
assumably fixed, closing here.

j-c-m:
Not fixed, certainly on freebsd, likely on all platforms.

On EAGAIN or EWOUlDBLOCK all platforms continue; the while, skipping any total / nwritten tallies, so subsequent loop, although now on a blocking socket, resends the same data.

j-c-m:
Here is a working patch for FreeBSD.

https://github.com/j-c-m/samba/commits/master/source3/lib/sendfile.c

PR #77, opened 2017-02-13
autobuild.py: adjust 278c92126 to fix submodule retval

axs-gentoo:
commit 278c921 added a 'stat' check on test_tmpdir in cleanup(), but
that location (which is assigned to be a subdir of testbase) does not
necessarily exist -- more specifically, does NOT exist on ldb, talloc,
replace, and tevent.  Making the call to stat on test_tmpdir be non-fatal
allows the cleanup to finish and a successful return value on the
autobuild run.

### DIFF
--- a/script/autobuild.py
+++ b/script/autobuild.py
@@ -455,7 +455,7 @@ class buildlist(object):
 def cleanup():
     if options.nocleanup:
         return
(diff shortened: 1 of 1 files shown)


urisimchoni:
Pushed as cad23629ac48253e508fd9bead2bb79bfa7ee3b8

PR #78, opened 2017-02-20
util_net: Add TCP_FASTOPEN socket option

neheb:
For users who want to enable it.

Signed-off by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/lib/util/util_net.c
+++ b/lib/util/util_net.c
@@ -1038,6 +1038,9 @@ static const smb_socket_option socket_options[] = {
 #ifdef TCP_DEFER_ACCEPT
   {"TCP_DEFER_ACCEPT", IPPROTO_TCP, TCP_DEFER_ACCEPT, 0, OPT_INT},
 #endif
+#ifdef TCP_FASTOPEN
+  {"TCP_FASTOPEN", IPPROTO_TCP, TCP_FASTOPEN, 0, OPT_INT},
+#endif
   {NULL,0,0,0,0}};
 
 /****************************************************************************


urisimchoni:
Seems to me like TCP_FASTOPEN is only valid for some applications (depends on the application protocol, because it doesn't handle duplicated SYN packets), so putting it as a global option for all protocols doesn't seem right. I can't think of a specific Samba-related protocol where this would break things (the protocol usually have an authentication phase), but it still looks like the application code should be deciding whether it wants to enable TFO.

What actual benefit does TFO bring to Samba?

Thanks,
Uri

neheb:
Just a latency reduction. The socket option is not enough to realize it however. On Linux, the proper sysctl needs to be set as well.

urisimchoni:
Has it improved a real latency problem (given that in general, the connections are long-lived)? What was the setup?

My point about TFO is that the latency reduction comes at the cost of not being fully backwards compatible with TCP (a duplicated packet creates two connections for user space and feeds the first segment of data to user space).

Since this change has a slight chance of breaking things, I think we'll be a lot more lenient towards introducing it if it solves a real problem for someone, hence my question about "actual benefit" - I'm asking whether patching samba and enabling this option actually helped improving user experience or performance in a setup.

Thanks,
Uri.

neheb:
Hmm point well taken. Will close.

bjacke:
Uri: Are you sure, that duplicate packages due to FTO are a problem? I don't think so actually. Dup packages can happen if the FTO key changes or if TFO was used and was disabled again. But in any case those first packages will be rejected by the TCP stack and not end up in user space at all.

Apart from that I think that TFO is not very usefull for Samba as we have long-living TCP connections and a slightly reduced time of one network roundtrip does not make sense for us. A workaround for people who want to enable FTO server-site anyway: setting  net.ipv4.tcp_fastopen=1027 should enable TFO for all listening sockets unconditionally.

Side note: The client site of TFO is another story and more difficult to implement - and the current Linux client site TFO implmentation will probably change soon and be closer to the FTO API of Darwin. It does not make sense put energy into this for us.

urisimchoni:
My "concern" is:
https://tools.ietf.org/html/rfc7413#section-6.1

So yes, it's uncommon, and yes, probably all "our" protocols can handle replay because there's some initial handshake before something actually happens. But each protocol must be declared as "TFO-safe", and hence it's not right architecturally to enable something for all protocols.

PR #83, opened 2017-04-07
s3:lib/messages: Don't call sec_init() too late

weiss:
Ensure that `sec_initial_uid()` returns the correct value by calling `sec_init()` first.  Without this change, the `sec_initial_uid()` calls in question would always return 0, so things go wrong when the initial UID isn't 0:

    $ smbcontrol smbd reload-config
    could not init messaging context

### DIFF
--- a/source3/lib/messages.c
+++ b/source3/lib/messages.c
@@ -247,6 +247,8 @@ static NTSTATUS messaging_init_internal(TALLOC_CTX *mem_ctx,
 	const char *priv_path;
 	bool ok;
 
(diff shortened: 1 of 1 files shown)


abartlet:
I'm very sorry this was left in github for so long.

This looks like it was finally fixed by another contributor by e895b6cf4a7eb3d50d618a022be74db85975bf69 in master for Samba 4.9.

The bug for backporting to 4.7/4.8 is https://bugzilla.samba.org/show_bug.cgi?id=13368

PR #85, opened 2017-04-13
build: Fix a typo

ita1024:
The list of modules should only contain valid module names
and not single characters.


abartlet:
Fixed by similar commit 06cdf8d7c4402b52aa27631a5556dde0b22d9487 in master or Samba 4.7.

Sorry for the delay in handling this.

PR #87, opened 2017-06-24
docs-xml: Sort input file list

bmwiedemann:
because filesystems return entries in undeterministic order
and that ends up in index.xml and influences index.html
preventing reproducible builds of samba packages (e.g. for openSUSE)

See https://reproducible-builds.org/ for why this matters

Signed-off-by: Bernhard M. Wiedemann <bwiedemann@suse.de>

### DIFF
--- a/docs-xml/Makefile
+++ b/docs-xml/Makefile
@@ -8,7 +8,7 @@ include Makefile.settings
 
 # Docs to build
 MAIN_DOCS = $(patsubst %/index.xml,%,$(wildcard */index.xml))
-MANPAGES = $(wildcard $(MANPAGEDIR)/*.?.xml)
+MANPAGES = $(sort $(wildcard $(MANPAGEDIR)/*.?.xml))
 
 # Lists of files to process
 MANPAGES_PLUCKER = $(patsubst $(MANPAGEDIR)/%.xml,$(PLUCKERDIR)/%.pdb,$(MANPAGES))


abartlet:
Thanks.  If it isn't too much trouble can you re-send to samba-technical manually?  it looks great, but sadly the mailing list integration broke a while back and I need to fix it. 

The only other thing is:
 - why is OpenSUSE using the docs-xml makefile at all?  It really should just be used for the 'make dist' on our end, the normal waf build should build the docs by default.

bmwiedemann:
I now sent it to samba-technical with git send-email.
Don't know why our samba does build this way (I'm not the maintainer), but @ddiss is.

https://build.opensuse.org/package/view_file/openSUSE:Factory/samba/samba.spec line 1133 has
```bash
pushd docs-xml
autoconf && ./configure
XML_CATALOG_FILES="file:///etc/xml/catalog file://$(pwd)/build/catalog.xml" make htmlman
```

ddiss:
@abartlet : for openSUSE we build Samba directly from the git source, rather than the release tarball. This means that we need to handle the docs build. I've responded to the patch via the ML.

abartlet:
Sure, but is there something this is building that the main waf build doesn't build that you need, rather than just have historically got?  Do you really need HTML man pages?

The docs build via the makefiles is not as accurate, as the paths are not correctly set. 

If you build manpages from the waf build, we fill in the correct paths into the docs based on the configure output.  Sadly only man manpages are delivered.

ddiss:
We package the HTML man pages via a separate samba-docs package. We're also manually invoking "make manpages" for ctdb manpage generation.

The spec file and source tarball (generated from Git) can be found at:
https://build.opensuse.org/package/show/network:samba:STABLE/samba

bjacke:
was applied with eedebe2ef1b6ff92975289b6076f236650744621
Thanks!

PR #89, opened 2017-07-25
waf: disable-python - fix talloc wscript if bundling disabled

axs-gentoo:
The pytalloc-util dependency logic in lib/talloc/wscript on a
standalone build checks for pytalloc-util in a manner that will
fail if bundling is disabled, this causes issues on
--disable-python builds of ldb, tevent, and samba.

This patch restructures the logic to skip checks if python
is disabled, instead just setting the temporary state variable
'using_system_pytalloc_util' to False

Successfully tested patch on ldb-1.1.31 and above, tevent-0.9.33,
and samba-4.7_rc3

### DIFF
--- a/lib/talloc/wscript
+++ b/lib/talloc/wscript
@@ -74,18 +74,21 @@ def configure(conf):
                                      implied_deps='replace'):
             conf.define('USING_SYSTEM_TALLOC', 1)
 
-        using_system_pytalloc_util = True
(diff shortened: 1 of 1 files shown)


metze-samba:
This is in master and will be included in talloc 2.1.11

abartlet:
The commit is in master as 5b316a4c7aa830194bf254942daa5aef88a0ed17 and 2.1.11 was released, and this will also be in Samba 4.8.

PR #90, opened 2017-07-30
Default ./configure to have regedit build as false

newEntera:
TL;DR; Don't accept my pull request.
Having in mind you have --with-regedit and no --without-regedit option it would make more sense to have     opt.SAMBA3_ADD_OPTION('regedit', default=False) rather than default=None and then set it to True.
Having in mind the following code as well:
    conf.env.build_regedit = False
    if not Options.options.with_regedit == False:
        conf.PROCESS_SEPARATE_RULE('system_ncurses')
        if conf.CONFIG_SET('HAVE_NCURSES'):
            conf.env.build_regedit = True

### DIFF
--- a/source3/wscript
+++ b/source3/wscript
@@ -59,7 +59,7 @@ def set_options(opt):
 
     opt.SAMBA3_ADD_OPTION('cluster-support', default=False)
 
-    opt.SAMBA3_ADD_OPTION('regedit', default=None)
+    opt.SAMBA3_ADD_OPTION('regedit', default=False)
 
     opt.SAMBA3_ADD_OPTION('fake-kaserver',
                           help=("Include AFS fake-kaserver support"), default=False)


urisimchoni:
Actually AFAICT there **is** a --without-regedit option. The SAMBA3_ADD_OPTION function adds both a "with" option and a "without" option. Also tried this on master (well, not totally up-to-date master but I don't recall many changes in that area).

newEntera:
I see now that --without-regedit works.
What I have found is that:
$ ./configure --help|grep regedit
  --with-regedit
            Build with regedit support (default=auto)
But there is no --without-regedit in help if you can add that please. I'm not able to find how to do that.

urisimchoni:
The --with-foo flag also implies the existence of a --without-foo flag. This is common practice in build configuration systems (autoconf to name one), and is documented in https://wiki.samba.org/index.php/Build_Samba_from_Source#enable.2Fdisable.2Fwith.2Fwithout_Options

Listing both options for each option would double the help message.

Given that the current behavior is pretty standard and documented, and the down-side of adding to the help string, I don't support changing that.

Thanks,
Uri.

newEntera:
Okay I understand. But there are --without options in the help message and this makes me think there are no without options for the rest that aren't described.
What about putting the link to https://wiki.samba.org/index.php/Build_Samba_from_Source#enable.2Fdisable.2Fwith.2Fwithout_Options
in the help message :) ?
Anyway I see help messages are deprecated nowdays and nobody wants to support them. Closing this pullrequest, sorry for trouble.

PR #93, opened 2017-08-06
Added libsmbclient APIs for setting configuration file

puran157:
Added libsmbclient APIs for setting configuration file and Set function callback which is called to log
Samba logs


MilhouseVH:
Any reason why this was closed? The facility to set the config location through the client API would be most welcome in order to avoid hacks that involve [temporarily changing the $HOME environment](https://github.com/xbmc/xbmc/blob/master/xbmc/filesystem/SMBFile.cpp#L99-L101) so that an app can use it's "own" `smb.conf`.

puran157:
I accidentaly forked this out of v4-7-test instead of master.
Here is the one from master branch
https://github.com/samba-team/samba/pull/95.

MilhouseVH:
Thanks. Look forward to this merging in future!

PR #94, opened 2017-08-06
smbc_readdirplus implementation in libsmbclient.h which return file_info structure

puran157:
smbc_readdirplus API is similar to that of smbc_readdir, but it returns struct file_info structure which has more information, like size, mode, uid, gid, create-time, change-time, modified time, access time etc.

In existing implementation, a call to smbc_opendir returns all information about files in current directory from SMB server but smbc_readdir returns smbc_dirent structure and a repeated call should made to smbc_stat to get information about all files in current directory.

With this PR, I maintain an another double linked list of struct file_info structure and that gets filled up when smbc_opendir is called. smbc_readdirplus would return the file_info structure while iterating over the list

Signed-off-by: Puran Chand pchand@vmware.com


puran157:
Closing this as opened another with review comments feedback.
https://github.com/samba-team/samba/pull/140

PR #95, opened 2017-08-06
Apis for logging and configuration

puran157:
Added libsmbclient APIs for setting configuration file and Set function callback which is called to log
Samba logs

Signed-off-by: Puran Chand pchand@vmware.com


MilhouseVH:
You might need to send an email to the [samba-technical mailing list](https://lists.samba.org/archive/samba-technical/2017-August/date.html)... the [README](https://github.com/samba-team/samba/blob/master/README#L114) suggests this should happen automatically, but I'm not seeing anything yet (either broken, or maybe just delayed?)

puran157:
Sure, I will do that. Even I was wondering why there is no email to mailing list.

puran157:
This had multiple commits to fix few issues found during selftest.
Raised a new one with single commit
https://github.com/samba-team/samba/pull/99

PR #97, opened 2017-08-10
samba_backup: fix error handling, back up tdb files properly

timkent:
This patch includes the following changes:

- Back up TDB files in the same manner as LDB files.
- Stop tar returning error code 1 on changed files in the private dir
  so the if statement no longer needs to ignore it.

Signed-off-by: Tim Kent <tim@kent.id.au>

### DIFF
--- a/source4/scripting/bin/samba_backup
+++ b/source4/scripting/bin/samba_backup
@@ -15,6 +15,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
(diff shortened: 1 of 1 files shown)


Hacklin:
Why are you using `[l|t]`?  And not `[lt]`.

abartlet:
Thanks for trying to improve our locking here.  It certainly did need work. 

Samba 4.9 includes an overhauled backup tool that correctly backs up all TDB and LDB files and obtains the correct locks while doing so, which the old script, even with this patch, couldn't do. 

See ```samba-tool domain backup```

PR #98, opened 2017-08-15
vfs_retention: add this module to make a share read only with a special time period.

lidan-fnst:
Add vfs_retention module to make a share read only with a special time period.
Since the suggestion of mail "Re: about the worm feature".


douglasbagnall:
Please use https://gitlab.com/samba-team/samba/ or the samba-technical list.

PR #99, opened 2017-08-15
Added libsmbclient APIs for setting configuration file

puran157:
and Set callback function which is called to log
Samba logs


puran157:
Changed the PR to have single commit instead of multiple commits

puran157:
Separated this in 2 separate patches.
https://github.com/samba-team/samba/pull/113
https://github.com/samba-team/samba/pull/112

PR #102, opened 2017-09-20
libwbclient: skip duplicate primary group in returned sids array

iboukris:
When calling wbcCtxAuthenticateUser, the primary group currently
being referenced twice in returned wbcAuthUserInfo.sids array.
Fix this by skipping the second addition.

Signed-off-by: Isaac Boukris <iboukris@gmail.com>


iboukris:
Sure, I'll post on ML and follow up on gitlab accordingly.
Closing here.

PR #104, opened 2017-10-23
Update libsmb_server.c

papadp:
cli_shutdown() overwrites the wanted errno returned from ``status = cli_tree_connect_creds(c, share, "?????", creds);``

This is likely the bug reported here https://bugzilla.samba.org/show_bug.cgi?id=13050

As I was having experiencing the same behavior when using samba bindings for python.


abartlet:
The patch on that bug https://attachments.samba.org/attachment.cgi?id=13751 is slightly more extensive, we should probably go with that one. 

We do do really need however is a test so we don't regress on this again.

abartlet:
Sorry for the long delay in replying on the status here.

David's patch (from the bug) was merged into master as 7470b9b18af282a742929d3fc90f4be5520428a1 and the bug indicates backports to 4.7.

PR #106, opened 2017-11-07
Fix several issues when building with solaris studio (bug #12961)

zehdeh:
This fixes several issues when building with Solaris Studio, see https://bugzilla.samba.org/show_bug.cgi?id=12961

This patch was already provided by Jiri Sasek <Jiri.Sasek@Sun.COM> in the bug ticket - I just adapted it to the latest build, so all credit to him.


douglasbagnall:
Thanks @zehdeh.

We should close this because
1. we are forsaking github in favour of gitlab.
2. the pull request seems to picked up a large number of unrelated commits.
3. the upstream patch (samba bug 12961) no longer applies cleanly.

abartlet:
I agree. Closing.

PR #111, opened 2017-11-27
Fix typo and add missing vfs_zfsacl manpage

b-a-t:
Fix typo in the description of the "getwd cache" option. Add missing
vfs_zfsacl.8 man page to the list of the manpages. Replaced tabs with
spaces where applicable.

Signed-off-by: Timur I. Bakeyev <timur@iXsystems.com>

### DIFF
--- a/docs-xml/smbdotconf/tuning/getwdcache.xml
+++ b/docs-xml/smbdotconf/tuning/getwdcache.xml
@@ -6,7 +6,7 @@
     <para>This is a tuning option. When this is enabled a 
     caching algorithm will be used to reduce the time taken for getwd() 
     calls. This can have a significant impact on performance, especially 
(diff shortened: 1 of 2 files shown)


b-a-t:
Filenames were also reshuffled in the `sort` order for easier maintaining.

abartlet:
Can you split this up into two commits please?

Samba commits generally avoid doing more than one thing at once.  Metze's rule is that the word 'and' shouldn't be in a commit message, is easy to remember if not absolute.

b-a-t:
Sorry for leaving this open, Andrew, looks like those changes were committed in:

8034b88d4e771663c4b7b581fb6b1992c33d5d96
3e43f80ad1f0ce2f0a807da64d1f5e8bd6512526
e9e4cd4d2b6ae376cba90fdb56b9c9684f2dc492

PR #113, opened 2017-11-29
Added smbc_SetConfiguration which lets the user set the smb.conf for libsmbclient code

puran157:


source3/libsmb/libsmb_setget.c
```diff
@@ -98,6 +98,19 @@ smbc_setDebug(SMBCCTX *c, int debug)
 	TALLOC_FREE(frame);
 }
 
+/** set configuration file */
+void smbc_setConfiguration(SMBCCTX *c, const char* file)
```
ddiss: Please use an error return here. E.g. ret = -1, errno = ENOENT if the file doesn't exist.
Also, please use consistent style - const char *file

source3/libsmb/libsmb_setget.c
```diff
@@ -98,6 +98,19 @@ smbc_setDebug(SMBCCTX *c, int debug)
...
+{
+	if (lp_load_client(file))
```
ddiss: This will see the new "file" configuration appended / concatenated to the existing (smb.conf) config, right? In such a case the name "setConfiguration" might be a little misleading.

puran157: No, It will be overwriting the configuration with that of defined in new configuration 'file'. The configuration not defined in 'file' would stay unchanged.
With this any code using libsmbclient APIs can maintain its own copy of smb.conf instead of relying on global smb.conf (from /etc/samba/smb.conf or $HOME environment variable)

puran157:
About torture tests, I was thinking to have a temporary smb.conf with modified configuration and use that to update configuration, verify if configuration is reflected or not, undo that with that of default smb.conf and verify if configuration is undone or not. 

I just don't know where to put that smb.conf and which smb.conf should be used to restore configuration. (Restoring configuration is important because we don't want other torture tests to get impacted)

Any suggestion in this regard would be great.

source3/libsmb/libsmb_setget.c
ddiss: If you have the following in smb.conf:
```
[global]
  workgroup = SMBCONF
  log level = 10
```
and in custom.conf:
```
[global]
  workgroup = CUSTOM
```
then call
```
ctx = smbc_new_context();
smbc_init_context(ctx);
smbc_setConfiguration(ctx, "custom.conf");
```
"log level = 10" will still be configured following the smbc_setConfiguration() call, right?

ddiss:
your test plan sounds reasonable - selftest provisions a client config via write_clientconf (in selftest/selftest.pl), so you might be able to pass that path down to your test for restoration. I'd suggest raising this question directly on the samba-technical@lists.samba.org , as you'll reach a much wider audience.

source3/libsmb/libsmb_setget.c
puran157: Yes

puran157:
Addressed the review comments and updated the changes in new PR. Closing this
https://github.com/samba-team/samba/pull/119

PR #118, opened 2017-12-21
logline of account becoming NT_STATUS_ACCOUNT_LOCKED_OUT

kkplein:
As agreed with Andrew Bartlett, I am creating this pull request, to define DBGC_CLASS DBGC_AUTH
It would be great if this could be merged. (perhaps even to the 4.7 branch...) But probably I need to create another pull request for the 4.7 branch? (first time at this....)

### DIFF
--- a/auth/credentials/credentials_krb5.c
+++ b/auth/credentials/credentials_krb5.c
@@ -35,6 +35,9 @@
 #include "auth/kerberos/pac_utils.h"
 #include "param/param.h"
 
+#undef DBGC_CLASS
+#define DBGC_CLASS DBGC_AUTH
+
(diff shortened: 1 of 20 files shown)


kkplein:
Hi,

Done.

I thought using the github interface (site / aplication) would be easier 
than using straight cli git, but it seems *many* things (like editing a 
commit) can only be done with the git cli client. Makes me wonder why 
they have the Github Desktop client at all.

And why such simple things are not implemented in the github site. :-(

Anyway, done now.

Better?

abartlet:
Merged as d39664fc66dafaf5cce2b830033df52da4d81695 in master for Samba 4.8.  Thanks!

PR #122, opened 2018-01-11
Remove build system info (uname -a)

sathieu:
Preventing reproducible builds while adding minor benefit.

More information at <https://reproducible-builds.org/>.


sathieu:
A similar patch is in Debian since 2016:
https://salsa.debian.org/samba-team/samba/blob/70fa2753486c10ea47e81d0ee8f6f73829fb09d5/debian/patches/no_build_system.patch

PR #123, opened 2018-01-26
Update gpo.py

lausser:
This if will always fail because the string can't start with \\\\ and // at the same time

### DIFF
--- a/python/samba/netcmd/gpo.py
+++ b/python/samba/netcmd/gpo.py
@@ -221,7 +221,7 @@ def del_gpo_link(samdb, container_dn, gpo):
 
 def parse_unc(unc):
     '''Parse UNC string into a hostname, a service, and a filepath'''
-    if unc.startswith('\\\\') and unc.startswith('//'):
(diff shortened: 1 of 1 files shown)


douglasbagnall:
@lausser: Sorry we have left this so long. We are abandoning Github for Gitlab.

Thanks. This looks like an obvious improvement. Could you

1. add this to the commit message:
```
Signed-off-by: Gerhard Lausser <gerhard.lausser@consol.de>
```
2. ideally assert you have personal copyright or follow the DCO steps here https://www.samba.org/samba/devel/copyright-policy.html

and for extra points:

3. either raise on https://gitlab.com/samba-team/samba/ or send to the mailing list (subscribe via https://lists.samba.org/mailman/listinfo/samba-technical)

I realise this is a lot of fussing around for such a small patch. Sorry and thanks again.

douglasbagnall:
@lausser: we ended up fixing this a different way in 4bc481c5cb7d772f473171949a3d522788bb7dcf
so we will close this PR.

lausser:
Oh, surprise! I had forgotten to respond :-)

PR #124, opened 2018-01-30
Fix samba_dlz when named "reloads".

kristjanvalur:
Please see https://bugzilla.samba.org/show_bug.cgi?id=13214
Using a singleton instance makes samba_dlz break when named does a "reload".  Remove the singleton management and allow named to initialize a new samba_dlz_state, and destroy the old on reload.

bind does a "reload" of a dlz module by:

1.  Creating a new dlz instance state via dlz_create()
2.  Initializing it with dlz_configure()
3.  Deleting the old instance state via dlz_destroy().

Using a singleton instance would cause samba_dlz to fail after a reload.  I don't know the precise reason, here are two possibles:

- dlz_configure() on an already running instance would break it somehow.
- Returning the **same** pointer from dlz_create() as the one already running may confuse the internal bookkeeping in **bind**.  Notice how a new instance is created _before_ the old is destroyed.

Note:  This effectively undoes https://github.com/samba-team/samba/commit/34eab45cba6989b66d37c7eb74d97846014ec4eb

The issues with "state->samdb" went away with talloc 2.0.  state->samdb is already a talloc_reference() from the old 'state', and a talloc_free(state) will decrement the refcount.

### DIFF
--- a/source4/dns_server/dlz_bind9.c
+++ b/source4/dns_server/dlz_bind9.c
@@ -40,6 +40,16 @@
 #include "dlz_minimal.h"
 #include "dns_server/dnsserver_common.h"
 
+/* 
+ * maintaining a singleton will break the 'reload' functionality
+ * triggered by 'rndc reload' and 'service reload named' which
+ * is used e.g. by logrotate.  It can be enabled by defining the following
+ * preprocessor macro
+ */
+#ifndef DLZ_BIND9_USE_SINGLETON
+#define DLS_BIND9_USE_SINGLETON 0
+#endif
+
 struct b9_options {
 	const char *url;
 	const char *debug;
@@ -72,8 +82,10 @@ struct dlz_bind9_data {
 	dns_dlz_writeablezone_t *writeable_zone;
 };
 
+#if DLZ_BIND9_USE_SINGLETON
 static struct dlz_bind9_data *dlz_bind9_state = NULL;
 static int dlz_bind9_state_ref_count = 0;
+#endif
 
 static const char *zone_prefixes[] = {
 	"CN=MicrosoftDNS,DC=DomainDnsZones",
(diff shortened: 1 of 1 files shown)


kristjanvalur:
I'll try to make time for that over the holidays, I only have a production environment to test on.  However, I'm curious as to why the pull request wasn't just accepted.  I got no comment whatsoever, but it clearly fixed the issue.  See the explanation above.

abartlet:
> However, I'm curious as to why the pull request wasn't just accepted. I got no comment whatsoever, but it clearly fixed the issue. See the explanation above.

We are well aware of a number of patches meant to address this issue.  However as we could not reproduce the issue (significant time was spent), and because bind9 is not part of our automated test system, we can't just apply patches and hope. 

What we did do is write tests that try to emulate every possible bad behaviour BIND9 could force on our plugin, to ensure we are robust to that.  That is 8c754b4396d958cb153c792615769efb00ed4bbd.

We are moving away from GitHub to GitLab, so bugzilla is probably the best place to continue the conversation with any backtraces, and GitLab is the place to test out further patches.  I've CC'ed Aaron who did the testing and logging work there and so I'll close this ticket for now.

abartlet:
> I got no comment whatsoever

For this I do apologise.  While documented and encouraged as a public contribution method, GitHub never gained traction with the Samba Team as a whole.  Ironically, this last couple of weeks (as a concerted effort has been made to triage, accept or close the remaining pull requests) has seen more activity here than in the years previous.  

Thankfully I can report we are much more active and reactive on GitLab:  https://gitlab.com/samba-team/samba https://wiki.samba.org/index.php/Samba_CI_on_gitlab 

I hope between GitLab and Bugzilla we can still work together to reproduce and resolve this issue.

PR #127, opened 2018-02-16
Python3 ports gpo

dmulder:


catalyst-joe-guo:
After the copyright line added, this patch passes samba autobuild.
Could any one review and merge it, please?

frenzymadness:
Hello.

I did a quick review and I have two questions:
1. Why are you using global variables in the test? It's not necessary when you don't change variables content in a function/method. I tried it without the `global` and it works.
2. Shouldn't be `pyparam_util` defined as the dependency of libgpo with extended lib name? Example here: https://github.com/samba-team/samba/blob/da72c4018ed92feab28025f8c6c29894f4c8691c/lib/ldb-samba/wscript_build#L22

Thanks a lot for your work!

dmulder:
Andrew just merged this into master (noticed as I went to implement your suggested changes).

PR #132, opened 2018-02-23
Python3: Bulk literal octal number conversion

noelpower:
This patch converts literal octal numbers to a format that is acceptable in both python2 and python3.
e.g.

convert
           os.chmod(some_path, 0770)      # valid in python2 only
           os.chmod(some_path, 0o770)    # valid in python2 and python3

The commits follow the same pattern of patch grouping as a previous bulk conversion patch


noelpower:
going to submit a combined bulk conversion for the octal and exception changes

PR #134, opened 2018-02-26
samba-tool domain trust: fix trust compatibility to Windows Server 1709 and FreeIPA

abbra:
Two patches from this pull request attempt to fix compatibilities to Windows Server 1709 and FreeIPA.

FreeIPA does not implement netr_DsRGetDCNameEx2() in a way that can be used by `samba-tool`, so a DC search fails when running `samba-tool domain trust create`. Insteda, use netr_DsRGetDCNameEx2() with a remote server name to call own DC. This should cause our own DC to use CLDAP discovery which is supported by FreeIPA.

Windows Server 1709 disabled SMB1 by default, so one has to set `client ipc min protocol = SMB2` to get trust established. While this is a proper fix going forward, it makes sense to default to SMB2 internally when establishing LSA and Netlogon RPC connections even if `smb.conf` lacks the correct option and fall back to an older protocol only if smb2 fails. This is an approach already used by FreeIPA DC for few years.

Submitting via github to get a test build run.

### DIFF
--- a/python/samba/netcmd/domain.py
+++ b/python/samba/netcmd/domain.py
@@ -1662,6 +1662,7 @@ class DomainTrustCommand(Command):
         self.remote_server = None
         self.remote_binding_string = None
         self.remote_creds = None
+        self.protocol_options = ['smb2', '']
 
     def _uint32(self, v):
         return ctypes.c_uint32(v).value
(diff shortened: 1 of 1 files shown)


abartlet:
Why is this needed?  Don't we just use SMB2 by default or is that lost somewhere in the client lib?

abbra:
The way we create trusts (in samba-tool and in FreeIPA), we depend on the session key which will not be there if SMB connection is not encrypted. So establishing trust will fail otherwise already.

abartlet:
Merged as c390728819e73cefbf02e0d52d22805930f4c45b and 4422f7382aad3090cb959ade030a02bf4fef81ac (a different patch set but getting smb2 support for RPC connections) in master for Samba 4.10

PR #137, opened 2018-03-02
Parallel test for master

abartlet:
Pull the ad_dc and NT4 tests into parallel workers to make autobuild much faster.

This should also allow more tests to run on travis-ci, which is why I'm proposing the patch here as that will trigger such a run and allow debugging of that.

If more tests run on travis-ci those outside the Samba team trying to do Python work will have a more reliable testsuite.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,6 +11,14 @@ build_samba:
     # this one takes about 4 hours to finish
     - python script/autobuild.py samba            --verbose --tail --testbase /tmp/samba-testbase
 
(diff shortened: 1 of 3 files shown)


abartlet:
Sadly this pushes the samba-o3 branch over the 50min maximum for travis-ci (which is one goal of these patches, to make travis-ci a reasonable check for python3 patches in particular).

We probably need to put 'none' in a different (new?) environment and split up more of the ad_dc tests.

If someone wants to take this on, please do!

PR #138, opened 2018-03-02
Python3 misc bugfixes

noelpower:


noelpower:
Some patches for some c-lib modules recently converted. In
both cases string parameters are expected in python2 and python3 where I
believe in python3 these api(s) need to accept 'bytes'.

abartlet:
I think it should also take a string for the string form of the GUID.

noelpower:
I am not sure what you mean, do you mean you wish it to be able to
handle both str and byte in py3 as input params? If so isn't this just
trying to subvert what python3 tries to do (however painful it actually
is) to properly distinguishing bytes and strings. Additionally this
doesn't sound right since the c api that is wrapped is 
GUID_from_data_blob not GUID_from_string.  While the python2 usage of
string is handy it isn't  correct (from a python3 perspective)

catalyst-joe-guo:
This one is already merged, can be closed now.

abartlet:
Merged into master for Samba 4.9 as bb88292cee5289742b72ff0249f01d527920301f

PR #139, opened 2018-03-05
Parallel test for master

abartlet:
I finally have a set of tests that pass a on travis-ci, covering more than just quicktest of ad_dc.

This runs all the nt4_dc and nt4_member tests, the ad_dc tests and most importantly the 'none' environment, where the python unit tests are written.

I plan to tidy up the commits, so this is just a preview, but the purpose is to make it easy for folks outside the Samba Team to get comprehensive CI on proposed changes. 

As a side-effect the build is more parallel so will run faster even for the full build.


abartlet:
Merged as d955ab2239dbeab649e69520e36109ca2c49ab89 in master for Samba 4.9

PR #140, opened 2018-03-06
Added ReadDirPlus API in libsmbclient which returns detailed file properties

puran157:


puran157:
smbc_readdirplus API is similar to that of smbc_readdir, but it returns struct libsmb_file_info structure which has more information, like size, mode, uid, gid, create-time, change-time, modified time, access time etc.

In existing implementation, a call to smbc_opendir returns all information about files in current directory from SMB server but smbc_readdir returns smbc_dirent structure. A repeated call should be made to smbc_stat to get more information about all files in current directory and it turns out to be an additional network call to SMB Server.

With this PR, I maintain an another double linked list of struct libsmb_file_info structure and that gets filled up when smbc_opendir is called. smbc_readdirplus would return the file_info structure while iterating over the list.

puran157:
Merged hence  closing

PR #142, opened 2018-03-07
Significant point about veto files.

stilez:
It might be expected that this parameter inherits on a directory, or prevents traversal. It only prevents the viewing or access of a listed directory, not its contents. Worth a mention.


urisimchoni:
Hi,

Sorry for responding so late, Samba's web presence has moved to gitlab (and there's always the samba--technical mailing list).

We need a Signed-off-by: line in the commit message.

Can you re-post to samba-technical or as a merge request to Gitlab? (see the new README.md)

Thanks,
Uri.

PR #148, opened 2018-03-22
Make Samba autobuild more parallel

abartlet:
This patch series allows Samba autobuilds on sn-devel to run in about 3 hours. 

It also uses that same split-up to enable builds on Travis CI for GitHub (allowing much more comprehensive tests for new pull requests) and GitLab CI (once appropriate runners are configured). 

We also test more build combinations and include fixes for winbindd that made those tests excessively long. 

I've also compared the list of tests run between this and master and they are as expected
[test-list.diff.txt](https://github.com/samba-team/samba/files/1835777/test-list.diff.txt)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,7 +36,7 @@ build_samba_fileserver:
   stage: build
   tags:
     - docker
-    - shared
+    - private
   script:
     # this one takes about 1 hours to finish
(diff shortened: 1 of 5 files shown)


abartlet:
Merged into master for Samba 4.9 as d955ab2239dbeab649e69520e36109ca2c49ab89

PR #150, opened 2018-03-26
selftest/check_output: convert bytes to str for Python 3

catalyst-joe-guo:
`BlackboxTestCase.check_output` will return bytes since it uses
`subprocess.communicate` underneath.
Convert expected string result to bytes for comparing.
This will be compatible for both Python 2 & 3.


abartlet:
Included in #156 in master

PR #151, opened 2018-03-26
selftest/common: fix dsdb_Dn comparison for Python 3

catalyst-joe-guo:
In Python 3, the builtin `cmp` funtion was dropped. And the `__cmp__` magic
method in object is no longer honored, which is replaced by 6 new methods:
`__eq__, __ne__, __lt__, __le__, __gt__, __ge__`.

This caused `tests.CommonTests` failed with `py3_compatiable=True`.
Fixed by adding the above methods.


abartlet:
Included in #156 in master

PR #152, opened 2018-03-26
Misc py3 c modules

noelpower:
Remainder of the c-modules to be ported. I've enabled the existing tests that cover these modules tor py3 (also split one of the test to try and make it easier to review)


catalyst-joe-guo:
Hi Noel, test failed since test suite `samba.tests.samba3` doesn't exist any more, we should remove it from `selftest/tests.py`.

selftest/tests.py
```diff
@@ -65,7 +65,7 @@
 planpythontestsuite("none", "samba.tests.upgrade")
 planpythontestsuite("none", "samba.tests.core", py3_compatible=True)
 planpythontestsuite("none", "samba.tests.common")
-planpythontestsuite("none", "samba.tests.provision")
+planpythontestsuite("none", "samba.tests.provision", py3_compatible=True)
 planpythontestsuite("none", "samba.tests.password_quality")
 planpythontestsuite("none", "samba.tests.samba3")
```
catalyst-joe-guo: This line should be removed now.

python/samba/samba3/__init__.py
```diff
@@ -80,20 +79,20 @@ def __len__(self):
...
+        return [k.rstrip(b"\x00") for k in self.db if not k.startswith(REGISTRY_VALUE_PREFIX)]
 
     def subkeys(self, key):
         """Retrieve the subkeys for the specified key.
 
         :param key: Key path.
         :return: list with key names
         """
-        data = self.db.get("%s\x00" % key)
+        data = self.db.get(b"%s\x00" % key)
```
catalyst-joe-guo: % formatting for bytes is added in Python 3.5, formatting like this line will cause error in Python 3.4:
https://www.python.org/dev/peps/pep-0461/

selftest/tests.py
noelpower: that's annoying, this must have been lost  during last rebase (was fixed previously)

abartlet:
To test the smbd module, enable samba.tests.ntacl and samba.tests.posixacl

abartlet:
See https://gitlab.com/catalyst-samba/samba/commit/c3398f593c67fe2aba9e91868cefcdde313018ad for Joe's work trying to get the modules linking against the right thing.

Joe's current work is at 
https://gitlab.com/catalyst-samba/samba/commits/misc-py3-c-modules-joeg-fork
the CI failed for him, but on (seemingly) unrelated things.

python/samba/samba3/__init__.py
noelpower: well, I was testing against 3.6, wasn't aware of this. However our python3 support is new, this is new code, imo for python3 we should mandate the minumum version (to be something reasonable) that will make life easiest for both porting and making writing it easier to write python2/python3 compatible code. How far back in the python3 version stream will we have to support otherwise ?

besides changes like

-data = self.db.get("%s\x00" % key)
+data = self.db.get(b"%s\x00" % key)

now becoming

data = self.db.get("%s\x00" % key)
+data = self.db.get(key.encode('utf8') + '\x00')

being even uglier than before as the constructs get more complicated (and we are sure to have some of those) it will become even more error prone and harder to easily change. Just my 2c

douglasbagnall: Noel, it is probably best to discuss minimum versions on the list, because almost nobody sees these review comments.

As far as I can tell, the main blocker for >3.4 is the main autobuild machine which runs Ubuntu 14.04 with 3.4.

catalyst-joe-guo:
After fixing the bytes formatting and libname issue, this patch passed autobuild last night. I will create a new PR soon.

PS: Inline comments on github will not be sent to Samba email list, I will use the global comment box from now on.

catalyst-joe-guo:
After @noelpower 's new push with more py3 enabled testsuites, autobuild failed with a few more bytes/str related issues. I've fixed the ones I found so far, and a new autobuild is running via gitlab-ci. I guess there will still be failures/errors, but we are much closer to the target now!

noelpower:
it's unfortunate the ci isn't set for >= 3.5 so I never saw the errors
:-( I would have fixed them otherwise. Unfortunately I am vacation today
and not back 'till tuesday.

catalyst-joe-guo:
I've created another PR based on this one: https://github.com/samba-team/samba/pull/155

Autobuild passed, review appreciated!

abartlet:
Mostly included in #156 now in master

PR #153, opened 2018-03-27
selftest/hostconfig: fix ambiguous import for Python 3

catalyst-joe-guo:
Use relative path to import module within package and enable absolute
import.


abartlet:
Included in #156 now in master

PR #154, opened 2018-03-29
GPO fixes

dmulder:
There is NO new functionality here, just some minor fixes and
reorganization of sources. The reorganization makes it possible to
inherit correctly and implement new gpos.


dmulder:
These patches are in the latest gpo pull request.

PR #157, opened 2018-04-05
Python2 python3 linkage issues rebase

noelpower:
This patch set ensures any py3 (or whatever is specified with --extra-python) built components like the correct libraries and objects, e.g. any py3 c-module built shouldn't directly (or indirectly) link in something that would cause both libpython2 and libpython3 to both linked. Additionally the grouping library (which doesn't seem to be be used) is removed.
In order to ensure the correct libraries are linked a couple of subsystems and libraries that weren't built for py3 (but should have been) are now built (so they can be linked)

### DIFF
--- a/auth/credentials/wscript_build
+++ b/auth/credentials/wscript_build
@@ -25,8 +25,11 @@ bld.SAMBA_SUBSYSTEM('CREDENTIALS_NTLM',
 	deps='samba-credentials')
 
 for env in bld.gen_python_environments():
+        pytalloc_util = bld.pyembed_libname('pytalloc-util')
+        pyparam_util = bld.pyembed_libname('pyparam_util')
+
 	bld.SAMBA_PYTHON('pycredentials',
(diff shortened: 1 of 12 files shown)


noelpower:
seems this is due to the way the build is split on the ci (with separate build/deliver) for ldb/talloc/tevent + core samba. I tried to reproduce this (by building in a similar way, previously I was just building everything from the main samba dir), I get a similar error (but not the same)
I fixed that error locally and am hoping the same fix will fix the error on the ci here

noelpower:
hmm failed again (but ci failures don't seem related), I am hoping rebasing against latest master which includes 567fe36c36c1d84e0edca151ece6a5b091995a49

noelpower:
This patch set now passes CI, it eliminates the incorrect simultaneous linking of both libpython2/libpython3 in python3 c - modules.

noelpower:
has been integrated via another pull request #163

PR #160, opened 2018-04-07
Fix samba_backup help about retpd

sathieu:


sathieu:
Thanks @abartlet , I found this little bug while reading https://bugs.debian.org/742177. I won't invest more time on it then (I presume the integration in samba-tool will be in 4.9 ?).

abartlet:
Yes, that is the plan.

sathieu:
Abandoning this change.

PR #161, opened 2018-04-10
WIP Py2/Py3 porting (not for review)

noelpower:
Changes to allow samba.tests.dsdb_lock is py2/py3 compatible. This tests a bit more some of the previous modules already ported


noelpower:
gah, seems like future module is not available from the ci machine :/ I think six, future etc. are essential to write python2/python3 compatible code. Otherwise we will just be reinventing the wheel here.

abartlet:
Indeed, Samba has chosen to reinvent the wheel with a smaller python2/3 layer.  There is a text_type in python/samba/compat.py that may be of assistance.

abbra:
six is essential but 'future' can be omitted. In this case you can replace future.iteritems by a native key iteration which will work in both py2 and py3:
```diff
-        for key, val in global_settings.iteritems():
+        for key in global_settings:
             f.write("\t%s = %s\n" % (key, global_settings[key]))
         f.write("\n")
```
I'd also recommend following a Conservative Porting Guide that was developed when porting many Python applications, including majority of Samba code: http://portingguide.readthedocs.io/en/latest/index.html

noelpower:
thanks for the pointer to the guide (trying to wrap my head around it)
hadn't come across this one

noelpower:
Changing the name of this patch set, using this branch for visibility/testing purposes. I'll try to keep this branch up to date (and will rebase as necessary) with current WIP patches from my own work.
Currently I am in the process of porting individual tests, I am trying to identify wider and more generic changes from the changes necessary to get the tests to run, as I do that I am pulling those changes out into separate pull requests)

PR #162, opened 2018-04-11
Check "auth event notification" param in log_json

reqa:
When Samba is compiled with JSON support it logs an error message
at log level 3 by default for every authentication event:

get_auth_event_server: Failed to find 'auth_event' registered on the
message bus to send JSON authentication events to:
NT_STATUS_OBJECT_NAME_NOT_FOUND

This patch adjusts the log_json function to consider the smb.conf
parameter "auth event notification" before calling auth_message_send.

Signed-off-by: Arvid Requate <requate@univention.de>

### DIFF
--- a/auth/auth_log.c
+++ b/auth/auth_log.c
@@ -201,6 +201,7 @@ static void auth_message_send(struct imessaging_context *msg_ctx,
  *
  */
 static void log_json(struct imessaging_context *msg_ctx,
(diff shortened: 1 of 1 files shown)


abartlet:
Merged as 228d0430d59d26ae12d1f2ab7082f86fa10d5ba0 in master for Samba 4.9.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=13381

PR #163, opened 2018-04-12
Python 3 porting patches part 1

catalyst-joe-guo:
I am still in progress of porting python selftest one by one, this is the job I've done so far. More PRs are coming soon.


python/samba/provision/__init__.py
```diff
@@ -37,11 +37,15 @@
...
-import urllib
 import string
 import tempfile
 import samba.dsdb
 
+try:
+    from urllib.parse import quote  # py3
+except ImportError:
+    from urllib import quote  # fall back to py2
+
```
noelpower: can't we use six here  (which will hide the try/except handling)
e.g.
    from six.moves.urllib.parse import quote
    
 or if not and we want to reinvent then a wrapping function (hiding the try/except) should go in python/samba/compat.py

python/samba/dbchecker.py
```diff
@@ -2306,7 +2307,7 @@ def check_object(self, dn, attrs=['*']):
                 self.fix_metadata(obj, att)
 
         if self.is_fsmo_role(dn):
-            if "fSMORoleOwner" not in obj and ("*" in attrs or "fsmoroleowner" in map(str.lower, attrs)):
+            if "fSMORoleOwner" not in obj and ("*" in attrs or "fsmoroleowner" in attrs_lower):
                 self.err_no_fsmoRoleOwner(obj)
                 error_count += 1
```
noelpower: hmm, while I don't disaggree with this change, AFAICS the change (list/map) for dbchecker,py above actually isn't necessary for py2/py3. It looks to me this code will run fine as is (under python2/python3) If I am correct then this change probably should be in a separate patchset right?

catalyst-joe-guo: My reasons:
1. don't repeat yourself
2. use `in` on a map iter has a risk:
```
x = map(str, range(1,2,3))
'1' in x   # first time, return True
'1' in x   # second time, return False
```
In this case it won't happen, but the new code is more robust.

noelpower: like I said, I don't have a problem with the change, It should however be in a separate commit (and/or patchset) as it isn't really a py2/py3 compatibility change (and isn't necesary for the py3 porting) It shouldn't be intermingled with the py3 porting changes, this change is imho independent (and yes... useful/good/etc)

catalyst-joe-guo: OK, removed.

catalyst-joe-guo:
Most of the commits are merged, close.

PR #164, opened 2018-04-12
fix ldb-samba: require pid match for cached ldb

dmulder:
Build failing with unknown function getpid() error

Signed-off-by: David Mulder <dmulder@suse.com>


abartlet:
Can we get some more detail on this?  I figure if we had a continuing issue here I would have heard more noise, but anything is possible I suppose.

abartlet:
Closing this because we don't use GitHub, praticularly for pull requests any more and this isn't the right fix in any case (we include system headers via the system/ headers from libreplace).

PR #165, opened 2018-04-13
traffic_relay: bulk port print to modern py3 style

catalyst-joe-guo:
Change print to function and avoid the ugly `print >>sys.stderr`.


catalyst-joe-guo:
Merged, close.

PR #166, opened 2018-04-16
Py3 string arg handling

noelpower:
Separate out patches from pull request #161 
Many of the python c-modules don't handle unicode as string types when passed say as object. This wasn't really a problem in a python2 only world. With code ported to be python2/python3 compatible it seems (at least in code we are porting) that we now find objects that were previously string in python2 now getting passed as unicode types in python2. This is a side affect of trying to cater for variables returned from functions as 'bytes'  in python3 that are returned as 'string' in python2. We normally do something like

   py2py3compat_var.decode('utf8')

To ensure the bytes provided in python3 are converted to 'str' The same line of code when running under python2 will covert a 'str' type to 'unicode'

We really should handle 'unicode' anyway as it is a 'text type. For example python2 code that use the ParseTuplexxxx functions to handle arguments already automatically handle 'str' or 
'unicode' when passed a 'string' format.

### DIFF
--- a/lib/ldb/pyldb.c
+++ b/lib/ldb/pyldb.c
@@ -1078,7 +1078,7 @@ static const char **PyList_AsStrList(TALLOC_CTX *mem_ctx, PyObject *list,
 		const char *str = NULL;
 		Py_ssize_t size;
 		PyObject *item = PyList_GetItem(list, i);
-		if (!PyStr_Check(item)) {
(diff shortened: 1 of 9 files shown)


noelpower:
integrated with merge of #171

abartlet:
This looks OK to me.  

Please push!

abartlet:
Closing as it landed in master with #171 in ba5f00deb772b76719bfd06467a6b05edfeef061

PR #169, opened 2018-04-23
fix traffic replay against windows

catalyst-joe-guo:
A few issues I founded while doing the performance test against windows with traffic replay.


abartlet:
This looks good to me.  

PR #175, opened 2018-05-04
Fix several spelling erroes

sathieu:
Caught by lintian.


sathieu:
Remaining spelling errors:
- dont don't (hard to match only strings and not code)
- ot to (ditto)
- prefered preferred (wanted synonym of `preferred master`)
- reponse response (`lm_reponse`, `nt_reponse`, is this intented?)
- ressource resource (touches code, and vfs_fruit param)
- upto up to (hard to match only strings and not code)
- wont won't (hard to match only strings and not code)

DHowett:
`ressource` is a dangerous one to change. For historical reasons, it was briefly a supported spelling for that config option and it needed to be retained for compatibility.

https://www.samba.org/samba/history/samba-4.6.0.html

abartlet:
The Heimdal changes need to go to Heimdal first, otherwise these look fine. 

sathieu:
@abartlet Thanks Andrew. You have skipped https://github.com/samba-team/samba/pull/175/commits/d5b54c72ba8a9a8b94748a90013adfc6d8d7b1ae `Fix spelling s/explicite/explicit/`

PR #176, opened 2018-05-06
More py2 py3 compat

noelpower:
Some general py2/py3 compatibility fixes.
* b64encode results in bytes in python3 but we invariably treat the result in our code as string. This changes converts all usage of b64encode to decode the result, this will result in unicode in python2. Treating text as a text type (e.g. in python2) should ensure some sanity down the line (and should expose anywhere we don't really treat the result as a string)
* keys() now returns 'dict' class not a list, enclose result of .keys with list() function where it seems that makes sense
* xrange -> range Note: there is one instance of this where it might make a difference (see commit message) due to the potential amount of items in the range. Advice needed here, might need to drop one of these patches at the moment. Six apparently has a wrapper we could use instead (but CI doesn't support yet... I think) 
* md5 takes bytes in python3, some patches to make sure this is the case
* StringIO has moved to 'io' in python3, use the compat wrapper
* some missed exception compatibility fixes
* .next() method not longer exists in python3

### DIFF
--- a/auth/credentials/tests/bind.py
+++ b/auth/credentials/tests/bind.py
@@ -96,7 +96,7 @@ operatingSystem: Windows Server 2003
 dn: """ + self.computer_dn + """
 changetype: modify
 replace: unicodePwd
-unicodePwd:: """ + base64.b64encode("\"P@ssw0rd\"".encode('utf-16-le')) + """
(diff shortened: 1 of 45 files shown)


abartlet:
Closing in favour of #185 (in autobuild) and #186.

PR #177, opened 2018-05-07
WIP: Make Group Policy Extensible, etc

dmulder:
I'm posting these for comment. One thing still to do is to load the registry with the default gp_sec_ext extension (otherwise it must be loaded manually before anything is applied). Unless anyone thinks this *should* be manually configured.

Machine extensions can be applied with:
sudo /usr/sbin/gpupdate --target=Computer

User extensions are applied with:
sudo /usr/sbin/gpupdate --target=User --username=Administrator --password=<password>

Machine extensions can also be applied via winbind (already).

This patchset includes:

- The check_refresh_gpo_list() patches only because it depends on them (these are being discussed in a different thread).
- A fix for the gp_sec_ext. The extension should check for itself whether it should apply, not the gpupdate script.
- Add libgpo bindings register/unregister/list gp_extensions in the registry.
- Dynamically load extensions registered in HKLM/Software/Microsoft/Windows NT/CurrentVersion/Winlogon/GPExtensions.
- Adds user policy extensions (this is now a trivial change).
- Renames samba_gpoupdate to gpupdate, and changes the --machine parameter to --targer=Computer/User. This is how the gpupdate script works in Windows.
- Fixes a crash in the gpupdate script when the dc code is installed, but the machine is joined as a client.

I'm posting these now for comments/suggestions/improvements.


abartlet:
I realise this is closed by do check with garming before pulling the gp_ext stuff as he is doing GPO parsing for a client job and is interested in re-animating some of this.

PR #178, opened 2018-05-08
python: Add argparse samba options

dmulder:
Creates samba.getarg, which mirrors the options in samba.getopt, accept these options work with the newer python argparse module (since optparse has been deprecated for some time).

import argparse
from samba import getarg as options
parser = argparse.ArgumentParser('program')
sambaopts = options.SambaOptions(parser)
credopts = options.CredentialsOptions(parser)
options.add_argument_group(parser, sambaopts)
options.add_argument_group(parser, credopts)
options.add_argument_group(parser, options.VersionOptions(parser))

There is nothing like optparse.add_option_group() anymore, argparse.add_argument_group() only accepts the parameters to the _ArgumentGroup class, instead of an instance of _ArgumentGroup. This is the reason for options.add_argument_group().
Also, the only way to write a callback is to implement an Action in argparse, hence the store() and store_true() generators.

### DIFF
--- a/python/samba/getarg.py
+++ b/python/samba/getarg.py
@@ -1,0 +1,313 @@
+# Samba-specific bits for argparse
+# Copyright (C) Jelmer Vernooij <jelmer@samba.org> 2007
+# Copyright (C) David Mulder <dmulder@suse.com> 2018
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


abartlet:
As discussed on the list, sadly argparse is not available in Python 2.6, which for $REASONS we must still support for now.

Please re-raise this as a GitLab merge request when we have gone pure Py3.

PR #182, opened 2018-05-10
traffic: optimize packet init for better performance

catalyst-joe-guo:
When we run traffic_replay, we are creating millions of Packet objects.
So small change in Packet.__init__ will make big difference.

By initializing packet with converted values without parsing string, the time
cost for 3961148 calls of Packet.__init__ dcrease from 17s to 4s, according
to cProfile.


abartlet:
merged into master as 21c82072ab87e3dee617b3219364e55e9c106432 for Samba 4.9

PR #183, opened 2018-05-10
traffic_replay: fetch domain from creds other than opts

catalyst-joe-guo:
For traffic_replay script, when user provides `--workgroup` or `-W` option
from command line, it will be set on the creds option group, other than the
opts one. So while checking this option, we should check it against creds.

The previous code is actually loading domain from lp, which get data from the
smb.conf file.


abartlet:
Merged into master for Samba 4.9 as 7181af9f9459d65df9df59ed447729e1c13a958c

PR #184, opened 2018-05-11
traffic: uniform stats output

catalyst-joe-guo:
The original code is trying to output different data format for tty or file.
This is unnecessary and cause confusion while writing script to parse result.

The human-readable one is also easy for code to parse.
Remove if check for isatty(), just make output the same.


abartlet:
This was merged into master as 68c64c634ad7576b2ab5ef838c203320afd90f44 for Samba 4.10

PR #185, opened 2018-05-11
py2/py3 patches reviewed

abartlet:
These are the py2/py3 patches I'm happy to have my review on.

See also:

https://gitlab.com/catalyst-samba/samba/pipelines/21849494
https://gitlab.com/catalyst-samba/samba/commits/abartlet-np-py3

### DIFF
--- a/auth/credentials/tests/bind.py
+++ b/auth/credentials/tests/bind.py
@@ -96,7 +96,7 @@ operatingSystem: Windows Server 2003
 dn: """ + self.computer_dn + """
 changetype: modify
 replace: unicodePwd
-unicodePwd:: """ + base64.b64encode("\"P@ssw0rd\"".encode('utf-16-le')) + """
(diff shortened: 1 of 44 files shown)


abartlet:
This is now in autobuild, except for the top patch around the DNS high byte (0xff).  I want to think on that some more, we really want to force a high byte into the end record, so I'll check on this more with @GSam tomorrow.  If a genuinely valid UTF8 sequence with a high byte will satisfy then we may wish to use that.

abartlet:
The Cstring patch needs to be after or alongside the DNS patch I dropped, or else the build fails.

In general try it is best if the patches all pass make test (in theory, there isn't an automated check for this part) to aid a git bisect or dropping patches from the series like I have.

noelpower:
I don't believe this has anything to do with cString, looking into it

PR #186, opened 2018-05-11
python/samba/netcmd: Py2/Py3 misc replace range with enumerate

noelpower:
Rework of patch from #176
Replace various instances of xrange with enumerate.

Signed-off-by: Noel Power <noel.power@suse.com>


noelpower:
merged with a52d0c3a14729b923f5488c67255ab069791b268

PR #187, opened 2018-05-14
Bso 13435

noelpower:
Fix for https://bugzilla.samba.org/show_bug.cgi?id=13435


noelpower:
merged with 83bde8a49cf3015c71302fbd209f95006e7535b2

PR #189, opened 2018-05-17
GPO Extensible

dmulder:
Makes GPO extensible via the registry key HKLM/Software/Microsoft/Windows NT/CurrentVersion/Winlogon/GPExtensions.
Adds user policies.
Modifies client side extensions to look and operate like they do on Windows.
Moves samba_gpoupdate to gpupdate, and updates options to operate like this command does in windows.
Adds offline apply.
Fixes deleted GPO issue, where settings from deleted GPOs were not being removed.
Other GPO fixes for issues found while creating these changes.


dmulder:
These tests stalled, not failed. Anybody know why (seems to be unrelated to my changes)?

PR #190, opened 2018-05-18
Make ldb configuration --disable-python work

b-a-t:
Fix broken in ldb 1.3 --disable-python configuration option.


abartlet:
Fixed in 95c117ff1114122aad367adab6c738b835a7c3d3 in master for ldb 1.4.0 and Samba 4.9

PR #193, opened 2018-06-06
samba-tool:demote: Remove dns references after demote a DC

catalyst-joe-guo:
In `samba-tool domain demote` command, call `remove_dc.remove_dns_references` after demote a DC.
Convert the `test_demote.sh` bash script to a Python unit test file, test demote command, and verify DNS records are removed as expected.


abartlet:
Merged as 0fb122af5094d568c56b0a7814696f3389813bbd in master for Samba 4.9

PR #196, opened 2018-06-13
Cached Group Policy and client application

dmulder:
On a client machine, we need to cache group policies before applying them. This also makes forcing policy possible when offline.
The cache path is determined using the lp.cach_path() function, which returns a cache directory within samba's cache, guaranteeing the cache has been created with all the correct permissions. The Group Policy Template (the gpo cache) is then downloaded to that cache using 0744 perms for the subdirs.
Client machines cannot read the GPT.INI and get the gpo versions without this patch, since they don't have access to the sysvol share locally (gpo_get_sysvol_gpt_version() reads from a local file).

There is also a test included which verifies that the cache download works.

### DIFF
--- a/python/samba/gpclass.py
+++ b/python/samba/gpclass.py
@@ -305,32 +305,16 @@ class gp_ext(object):
     def read(self, policy):
         pass
 
-    def parse(self, afile, ldb, conn, gp_db, lp):
+    def parse(self, afile, ldb, gp_db, lp):
         self.ldb = ldb
         self.gp_db = gp_db
         self.lp = lp
 
-        # Fixing the bug where only some Linux Boxes capitalize MACHINE
-        try:
-            blist = afile.split('/')
-            idx = afile.lower().split('/').index('machine')
-            for case in [
-                            blist[idx].upper(),
-                            blist[idx].capitalize(),
-                            blist[idx].lower()
-                        ]:
-                bfile = '/'.join(blist[:idx]) + '/' + case + '/' + \
-                    '/'.join(blist[idx+1:])
-                try:
-                    return self.read(conn.loadfile(bfile.replace('/', '\\')))
-                except NTSTATUSError:
-                    continue
-        except ValueError:
-            try:
-                return self.read(conn.loadfile(afile.replace('/', '\\')))
-            except Exception as e:
-                self.logger.error(str(e))
-                return None
+        local_path = self.lp.cache_path('gpo_cache')
+        data_file = os.path.join(local_path, afile).replace('\\', '/')
+        if os.path.exists(data_file):
+            return self.read(open(data_file, 'r').read())
+        return None
 
     @abstractmethod
     def __str__(self):
@@ -466,11 +450,6 @@ def gpo_version(lp, path, sysvol):
 def apply_gp(lp, creds, test_ldb, logger, store, gp_extensions):
     gp_db = store.get_gplog(creds.get_username())
     dc_hostname = get_dc_hostname(creds, lp)
-    try:
-        conn =  smb.SMB(dc_hostname, 'sysvol', lp=lp, creds=creds)
-    except:
-        logger.error('Error connecting to \'%s\' using SMB' % dc_hostname)
-        raise
     gpos = get_gpo_list(dc_hostname, creds, lp)
     try:
         check_refresh_gpo_list(dc_hostname, lp, creds, gpos)
@@ -494,7 +473,7 @@ def apply_gp(lp, creds, test_ldb, logger, store, gp_extensions):
         store.start()
         for ext in gp_extensions:
             try:
-                ext.parse(ext.list(path), test_ldb, conn, gp_db, lp)
+                ext.parse(ext.list(path), test_ldb, gp_db, lp)
             except Exception as e:
                 logger.error('Failed to parse gpo %s for extension %s' % \
                     (guid, str(ext)))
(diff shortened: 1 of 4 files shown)


python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
...
+                f.write(conn.loadfile(fname))
+
+def check_refresh_gpo_list(dc_hostname, lp, creds, gpos):
+    conn = smb.SMB(dc_hostname, 'sysvol', lp=lp, creds=creds)
+    cache_path = lp.cache_path('gpo_cache')
+    for gpo in gpos:
+        if not gpo.file_sys_path:
+            continue
+        dirs = gpo.file_sys_path.split('\\')
+        dirs = dirs[dirs.index('sysvol')+1:]
```
dmulder: Make sure sysvol is in the path

python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
         gpos = ads.get_gpo_list(creds.get_username())
     return gpos
 
+FILE_ATTRIBUTE_DIRECTORY = 0x10
+def cache_gpo_dir(conn, cache, sub_dir):
+    try:
+        os.makedirs(os.path.join(cache, sub_dir), mode=0o744)
```
dmulder: should be 755

python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
...
+    for fdata in conn.list(sub_dir):
+        if fdata['attrib'] & FILE_ATTRIBUTE_DIRECTORY:
+            cache_gpo_dir(conn, cache, os.path.join(sub_dir, fdata['name']))
+        else:
+            with open(os.path.join(cache, sub_dir, fdata['name']), 'w') as f:
+                fname = os.path.join(sub_dir, fdata['name']).replace('/', '\\')
+                f.write(conn.loadfile(fname))
+
+def check_refresh_gpo_list(dc_hostname, lp, creds, gpos):
+    conn = smb.SMB(dc_hostname, 'sysvol', lp=lp, creds=creds)
```
dmulder: ensure this connection signed

python/samba/gpclass.py
```diff
@@ -319,18 +324,18 @@ def parse(self, afile, ldb, conn, gp_db, lp):
                             blist[idx].capitalize(),
                             blist[idx].lower()
```
dmulder: force lower case when downloading the files

python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
...
+        pass # File Exists
+    for fdata in conn.list(sub_dir):
+        if fdata['attrib'] & FILE_ATTRIBUTE_DIRECTORY:
+            cache_gpo_dir(conn, cache, os.path.join(sub_dir, fdata['name']))
+        else:
+            with open(os.path.join(cache, sub_dir, fdata['name']), 'w') as f:
+                fname = os.path.join(sub_dir, fdata['name']).replace('/', '\\')
+                f.write(conn.loadfile(fname))
+
+def check_refresh_gpo_list(dc_hostname, lp, creds, gpos):
```
dmulder: ensure gpos is coming across signed ldap in the capi

python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
...
+    conn = smb.SMB(dc_hostname, 'sysvol', lp=lp, creds=creds)
+    cache_path = lp.cache_path('gpo_cache')
+    for gpo in gpos:
+        if not gpo.file_sys_path:
+            continue
+        dirs = gpo.file_sys_path.split('\\')
```
dmulder: check for / slashes also, just in case

dmulder: also check for '..'

python/samba/gpclass.py
```diff
@@ -423,23 +428,59 @@ def get_gpo_list(dc_hostname, creds, lp):
...
+        dirs = dirs[dirs.index('sysvol')+1:]
+        cache_gpo_dir(conn, cache_path, os.path.join(*dirs))
+
+def gpo_version(lp, path, sysvol):
+    # gpo.gpo_get_sysvol_gpt_version() reads the GPT.INI from a local file.
+    # If we don't have a sysvol path locally (if we're not a kdc), then
+    # read from the gpo client cache.
+    if sysvol:
+        local_path = os.path.join(sysvol, path, 'GPT.INI')
```
dmulder: don't maintain two code paths, just include the gpt cache on kdc also

python/samba/tests/gpo.py
```diff
@@ -75,3 +76,17 @@ def test_gpt_version(self):
...
+        if ads.connect():
+            gpos = ads.get_gpo_list(self.creds.get_username())
+        check_refresh_gpo_list(self.server, self.lp, self.creds, gpos)
+
+        assert os.path.exists(cache), 'GPO cache %s was not created' % cache
+
+        guid = '{31B2F340-016D-11D2-945F-00C04FB984F9}'
+        gpt_ini = os.path.join(cache, 'addom.samba.example.com/Policies',
+                               guid, 'GPT.INI')
+        assert os.path.exists(gpt_ini), 'GPT.INI was not cached for %s' % guid
```
dmulder: test for malicious file paths

PR #197, opened 2018-06-13
Make Group Policy Exensible

dmulder:
Trying this a little different way now (third set of patches for this feature).
This works *mostly* the way it's done on a Windows client.
Normally on a windows client, Group Policy client side extensions are registered via the Windows Registry in designated key, but Andrew pointed out that this is unreliable in samba. So I've moved the registered extensions to a configuration file in the cache path. Another advantage to this approach is that it can all be written in python code now (so we don't have to deal with the python c-api).
I've also added input validation so we don't end up with garbage in the gpext config file.
Testing is also included.

See the MS spec here regarding how CSE registration works (under Creating and Registering):
https://msdn.microsoft.com/en-us/library/aa374288(v=vs.85).aspx

### DIFF
--- a/python/samba/gpclass.py
+++ b/python/samba/gpclass.py
@@ -515,3 +515,31 @@ def register_gp_extension(guid, name, path,
     f.close()
 
     return True
+
+def unregister_gp_extension(guid, smb_conf=None):
+    # Check for valid guid
+    if not guid[0] == '{' or not guid[-1] == '}':
+        return False
+    try:
+        uuid = guid[1:-1]
+        UUID(uuid, version=4)
+    except:
+        return False
+
+    lp = LoadParm()
+    if smb_conf is not None:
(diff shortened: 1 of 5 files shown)


python/samba/gpclass.py
```diff
@@ -479,3 +486,86 @@ def unapply_gp(lp, creds, test_ldb, logger, store, gp_extensions):
...
+        UUID(uuid, version=4)
+    except:
+        return False
+
+    lp = LoadParm()
+    if smb_conf is not None:
+        lp.load(smb_conf)
+    else:
+        lp.load_default()
+    ext_conf = lp.cache_path('gpext.conf')
```
dmulder: this should be in state path, not cache

dmulder:
Moved to gitlab, with state path fix.

PR #200, opened 2018-06-25
MAN: Adding entry for net ads lookup

amitkumar50:
There is no man page description for net ads lookup.
This PR adds entry for the same.

### DIFF
--- a/docs-xml/manpages/net.8.xml
+++ b/docs-xml/manpages/net.8.xml
@@ -551,6 +551,12 @@ account in server manager first.</para>
 <para>Create specified group.</para>
 
 </refsect3>
+<refsect3>
+<title>[ADS] LOOKUP</title>
+
+<para>Lookup the closest Domain Controller in our domain and retrieve server information about it.</para>
+
+</refsect3>
 </refsect2>
 
 <refsect2>


cryptomilk:
Amit, thank you very for your contribution.

The patch requires your 'Signed-off-by', see https://www.samba.org/samba/devel/copyright-policy.html for all the details.

docs-xml/manpages/net.8.xml
```diff
@@ -550,6 +550,12 @@ account in server manager first.</para>
 
 <para>Create specified group.</para>
 
+</refsect3>
+<refsect3>
+<title>[ADS] LOOKUP</title>
+
+<para>Find the ADS DC using CLDAP lookup.</para>
```
cryptomilk: I would use:

Lookup the closest Domain Controller in our domain and retrieve server information about it.

cryptomilk:
Also, please send patches as an attachement to the samba-technical mailing list. Then they will be reviewed more quickly. Thanks!

docs-xml/manpages/net.8.xml
amitkumar50: @cryptomilk Thanks for comment.
But customers like to have sync between 'net help ads lookup' and 'man net' entries.
Though I agree exact duplicate of 'net help ads lookup' would not be good. But somehow I believe this would be good:
>Lookup the closest Domain Controller in our domain and retrieve server information for same using CLDAP lookup. 

What's your thoughts

amitkumar50:
Sure. So you means to do two things:
1. Open PR
2. Send patch as attachment to samba-technical

docs-xml/manpages/net.8.xml
cryptomilk: Use:

`<para>Lookup the closest Domain Controller in our domain and retrieve server information about it.</para>`

cryptomilk:
1. Send the Certificate of Origin -> https://www.samba.org/samba/devel/copyright-policy.html
2. Change the patch like discussed above
3. Add your Sign-Off -> git commit --amend -s
4. Send the patch as an attachment to the samba-technical mailinglist

abartlet:
BTW, the travis-ci fails because the patch set hasn't been rebased on master since I fixed the .travis-ci.yml file.

abartlet:
Merged as 56d9b1a533b8d3ebd4e686145883c7ed62c08ce0 in master for Samba 4.9

PR #202, opened 2018-07-04
[RFE] smbstatus -L does not resolve UIDs to names

amitkumar50:
This PR show username instead of UIDs in smbstatus -L output.

 /usr/local/samba/bin/smbstatus -L
Locked files:
Pid          Uid        DenyMode   Access      R/W        Oplock           SharePath   Name   Time
`--------------------------------------------------------------------------------------------------`
15577        fred       DENY_NONE  0x100081    RDONLY     NONE             /share   .   Wed Jul  4 17:27:17 2018
15577        fred       DENY_NONE  0x100081    RDONLY     NONE             /share   .   Wed Jul  4 17:27:17 2018

Signed-off-by: Amit Kumar amitkuma@redhat.com

### DIFF
--- a/source3/utils/status.c
+++ b/source3/utils/status.c
@@ -131,7 +131,7 @@ static int print_share_mode(const struct share_mode_entry *e,
 
 	if (count==0) {
 		d_printf("Locked files:\n");
-		d_printf("Pid          Uid        DenyMode   Access      R/W        Oplock           SharePath   Name   Time\n");
+		d_printf("Pid          Username        DenyMode   Access      R/W        Oplock           SharePath   Name   Time\n");
 		d_printf("--------------------------------------------------------------------------------------------------\n");
 	}
 	count++;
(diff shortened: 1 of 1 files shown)


slowfranklin:
Looks like smbstatus -L is specified to return numeric values by default, isn't it? Also, you add a call to uidtoname() which return const char * and then pass that to %d printf conversion specifier.

I guess there's a reason why we only print numeric values for the open file list: uidtoname() is expensive, so doing that with a large number of files may put severe load on the server.

amitkumar50:
@slowfranklin Thanks for review.
But I feel there is no use in showing UID to admin in this case.
Whole point is understanding locked files .. resource in one command.

cryptomilk:
@amitkumar50 by default, this command should not print names only numeric values. However what we could do is to add a new command line option `--resolveuids` which will then resolve the uids to usernames.

Do you want to code that?

douglasbagnall:
@amitkumar50, for this PR to go anywhere, you will need to the `--resolveuids` option mentioned by @cryptomilk, and submit it to https://gitlab.com/samba-team/samba. We do not use github.

PR #203, opened 2018-07-20
waf: libtdb: set version in dynlib on OSX (for Homebrew packaging)

alexeicolin:
This patchset is needed to build libtdb on OS X via Homebrew. To get the brew formula for libtdb into Homebrew core repository, Homebrew requires patches to be merged upstream.

On OSX, without this patch `-compatibility_version` and `-current-version` are
passed to clang during compilationg and not during linking. The resulting
libtdb.dynlib ends up without versions set (`otool -l` reports them as 0.0.0).
I don't think adding these to compile flags is useful at all, clang ignores
them. Not having version set is not fatal, but the current code already
attempts to set it -- this patch is a potential fix to existing behavior.

The rest of the patch is to remove these flags whenever `-dynamiclib` is
removed, because these flags only make sense with it. Unfortunately, this
removal happens in two places, due to pre-existing code duplication -- this
patch just changes both places.

Minor behaviour change, in case it has some unintended sideffects:
`-single_module` flag was only removed if `-dynamiclib` was present; after this
patch, they are both removed separately if present.

I only tested this patch against the *standalone* release of libtdb v1.13.6
from here: https://www.samba.org/ftp/tdb/. I built in the context of the brew
formula for tdb that I am about to submit to Homebrew.

Also, note that that I had to `./configure` with `--disable-rpath
--disable-rpath-install --disable-rpath-private-install`, otherwise
applications do link against the resulting libtdb.dynlib but fail at runtime,
because they are looking for libtdb.inst.dynlib instead of libtdb.dynlib and
not finding it.

Signed-off-by: Alexei Colin <ac@alexeicolin.com>

### DIFF
--- a/buildtools/wafsamba/wafsamba.py
+++ b/buildtools/wafsamba/wafsamba.py
@@ -1018,6 +1018,24 @@ def apply_bundle_remove_dynamiclib_patch(self):
         if not getattr(self,'vnum',None):
             try:
                 self.env['LINKFLAGS'].remove('-dynamiclib')
+
+                # Only try to remove these if -dynamiclib was preset, i.e. no exception above
+                try:
+                    arg_idx = self.env['LINKFLAGS'].index('-compatibility_version')
+                    self.env['LINKFLAGS'].pop(arg_idx) # the argument
+                    self.env['LINKFLAGS'].pop(arg_idx) # the subargument
+                except ValueError:
+                    pass
+                try:
+                    arg_idx = self.env['LINKFLAGS'].index('-current_version')
+                    self.env['LINKFLAGS'].pop(arg_idx) # the argument
+                    self.env['LINKFLAGS'].pop(arg_idx) # the subargument
+                except ValueError:
+                    pass
+            except ValueError:
+                pass
+
+	    try:
                 self.env['LINKFLAGS'].remove('-single_module')
             except ValueError:
                 pass
--- a/third_party/waf/wafadmin/Tools/gcc.py
+++ b/third_party/waf/wafadmin/Tools/gcc.py
@@ -94,8 +94,8 @@ def gcc_modifier_cygwin(conf):
 @conftest
 def gcc_modifier_darwin(conf):
 	v = conf.env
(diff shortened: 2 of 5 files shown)


alexeicolin:
The second commit fixes the error raised by Homebrew audit tool: Python extensions must not be linked against a specific Python library. Instead, should link to `-undefined dynamic_lookup`.

Both of the patches in this commit are required for the libtdb to be accepted into Homebrew core repository.

abartlet:
I realise this is difficult from your position, but we like patches to waf to first to go waf, then be ported into samba.  An effort is ongoing here:
https://gitlab.com/samba-team/devel/samba/commits/abbra-py3-master-waf

Could you try that branch and provide feedback to @abbra so he can perhaps find a way to do this without patching waf or work with @ita1024 on the upstream changes needed?

alexeicolin:
I tried that branch. It appears that the special case for OS X that I tried to remove in this PR is gone. The waf now uses `python-config --ldflags` to get the flags for linking, which makes sense to me. But the flags are not liked by the Homebrew auditor. I asked on their forum: https://discourse.brew.sh/t/audit-complains-about-python-extension-modules-being-linked-to-python-framework-binary-but-upstream-is-using-python-config/2688

The compatibility version is still set to 0.0.0 but this is not a blocker for packaging. If we do want to set it, we'd need to ask @abbra whether the new waf code attempts to set the compat and current version to anything and it's not working (like current waf code) or if it's not trying to set it and this would be a new feature to add to the new waf code.

I'll probably just close this since the new waf code from that branch will be eventually merged, making this patch on top of old waf obsolete no matter what.

abartlet:
I agree.  Let's revisit it once the new waf lands and you get an answer from the homebrew folks.

Finally, I'm told the py*-util libs (pyldb-util, pytalloc-util) are pretty special in the python world, so there isn't much good guidance on linking, symbol versions etc.  While I realise you are just doing tdb, I don't think assuming the python ABI is unchanging (so no need to link to it) is viable, particularly for python3 (where we are going soon).

PR #205, opened 2018-08-13
Add a lttng VFS module for samba.

thesues:
Tracing VFS functions using lttng is now possible

https://lttng.org/

Signed-off-by: Dongmao Zhang <deanraccoon@gmail.com>


douglasbagnall:
Samba does not use github. It looked like we did for a short time (sorry), but we don't. The correct place to submit this, once the merge conflicts are fixed is https://gitlab.com/samba-team/samba/

PR #206, opened 2018-08-24
Relax GPGME version check

sathieu:
The current method require at least the same version as during the build. This is not needed.

Used symbols indicate the minimum is 1.2.0

Bug-Debian: http://bugs.debian.org/886333


sathieu:
Closing in favor of https://gitlab.com/samba-team/samba/merge_requests/268.

PR #208, opened 2018-09-20
Improve amd64 detection on sunos

amalleo25:
SunOS may not include x86_64 or amd64 in uname output. Instead, isainfo can be used to determine architecture. SunOS also does not support noexecstack option (to my knowledge) so we skip noexecstack linking check.

Signed-off-by: Adam Malleo <amalleo25@gmail.com>

### DIFF
--- a/third_party/aesni-intel/wscript
+++ b/third_party/aesni-intel/wscript
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 from waflib import Options, Utils
-from subprocess import Popen, PIPE
 
 def configure(conf):
     if Options.options.accel_aes.lower() != "intelaesni":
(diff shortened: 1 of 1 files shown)


abartlet:
Please use Utils.cmd_output not Popen() directly

amalleo25:
Fixed

abartlet:
I'm sorry for the long delay in dealing with this.  To merge this into master this will need to work with python3, and I expect some bytes and strings issues (POpen and things that wrap it return bytes).

Also, it looks like this no longer applies on master.

Ideally you would re-open it on our GitLab instance, where we now do our web-based work, but if you could fix those issues and (most importantly) re-test on a SunOS system that would be most helpful.

PR #209, opened 2018-09-24
Use canonical path for Perl

sathieu:
Samba codebase mostly uses '#!/usr/bin/perl' as perl's shbang.

This ensures uniform Perl modules availibility.

### DIFF
--- a/examples/misc/wall.perl
+++ b/examples/misc/wall.perl
@@ -1,4 +1,4 @@
-#! /usr/bin/env perl
+#!/usr/bin/perl
 # 
 #@(#) smb-wall.pl Description:
 #@(#) A perl script which allows you to announce whatever you choose to
--- a/pidl/pidl
+++ b/pidl/pidl
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
 
 ###################################################
 # package to parse IDL files and generate code for
--- a/script/show_testsuite_time
+++ b/script/show_testsuite_time
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
 use Time::Local ('timegm');
 my $in = STDIN;
 use strict;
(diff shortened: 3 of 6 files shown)


bjacke:
just that most script use /usr/bin/perl instead of "/usr/bin/env perl" doesn't mean this this is the better thing to do. IMHO we should use "/usr/bin/env perl"  everywhere to support systems where perl is not installed in /usr/bin/

thctlo:
This is more a Debian specific setting, i agree with björn.

sathieu:
I think nowadays almost nobody (except perl devs) use perl installed on an exotic path.

bjacke:
there are, believe me :-)

sathieu:
I didn't expect opposition on this patch. Let's give some more remarks:
- Current codebase has 66 uses of `#!/usr/bin/perl` and 6 of `env.*perl`
- Nobody complained about those 66
- Using perl from `PATH` can break things (if a module is not found for example)
- Even a perl dev wants Samba to work, even if the /usr/local perl install is broken
- You can always call a script with `/usr/local/bin/perl script-name`

thctlo:
Yes, Mathieu, good points. Having 2 types of setting isnt right also.

bjacke:
On 24.09.2018 15:13, Mathieu Parent wrote:
> I didn't expect opposition on this patch. Let's give some more remarks:
>
>   * Current codebase has 66 uses of |#!/usr/bin/perl| and 6 of |env.*perl|
>   * Nobody complained about those 66

can you make a patch for those 66 then, please?

>   * Using perl from |PATH| can break things (if a module is not found
>     for example)

that's a problem of the perl installation then.

>   * Even a perl dev wants Samba to work, even if the /usr/local perl
>     install is broken

this is not about perl devs. There are systems, where perl is not part
of a standard installation and wherer perl in just not located in /usr/bin/

>   * You can always call a script with |/usr/local/bin/perl script-name|

this is fixing the non-portable shebangs. We want (at least try) to be
as portable as possible.

Please also use bugzilla to report bugs and sugget patches. Even if the
README still contains the outdated information about contributing
patches via github.

sathieu:
OK. Will do this patch next month.

PR #210, opened 2018-10-05
Don't use sysconf(_SC_NGROUPS_MAX) MacOS

arichardson:
On MacOS sysconf(_SC_NGROUPS_MAX) always returns 16. However, this is not
the value used by getgroups(2). MacOS uses nested groups but getgroups(2)
will return the flattened list which can easily exceed 16 groups. In my
testing getgroups() already returns 16 groups on a freshly installed
system. And on a 10.14 system the root user is in more than 16 groups by
default which makes it impossible to run smbd without this change.

See https://bugzilla.samba.org/show_bug.cgi?id=8773

### DIFF
--- a/source3/lib/system.c
+++ b/source3/lib/system.c
@@ -776,7 +776,18 @@ void sys_srandom(unsigned int seed)
 
 int groups_max(void)
 {
-#if defined(SYSCONF_SC_NGROUPS_MAX)
+#if defined(DARWINOS)
+	/*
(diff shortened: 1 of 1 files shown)


bjacke:
setting this fixed to 128 is definetely not right, a configure test is needed and the actual value should come from somewhere and not out of nothing. Please open a bugzilla bug for that and propose a patch there. Github is not used by Samba any more, even if the README file still contains that false informatin.

PR #211, opened 2018-10-08
Always use /usr/bin/env to call Perl

sathieu:
... to support systems where perl is not installed in /usr/bin/.


sathieu:
Replaces #209.

sathieu:
Closed in favour of https://bugzilla.samba.org/show_bug.cgi?id=13652

PR #213, opened 2018-11-23
Fix bind DLZ module to prevent zone transfers for everyone

littlejawa:
See https://bugzilla.samba.org/show_bug.cgi?id=9634 for bug reference and suggested changes.

- add an option to smb.conf to list authorized zone transfer clients
- implement restriction in dlz module to allow transfers only to selected IPs


littlejawa:
Hi @abartlet ,
I made a merge request on gitlab here: https://gitlab.com/samba-team/samba/merge_requests/169.

I can extend the test cases as needed, just not sure how - can we discuss it on the gitlab thread?

abartlet:
Yes

PR #214, opened 2018-11-26
Add --authentication-file option support to "net"

ojwb:
This patch extends net to support the same `-A authfile`/`--authentication-file authfile` option that most of the other tools already do.


ojwb:
I gather gitlab is now preferred to github, so I'll close this and open an MR there instead.

PR #215, opened 2018-11-30
Send status to systemd on daemon start

marcosfrm:
Small cosmetic fix.

https://bugzilla.samba.org/show_bug.cgi?id=11568
https://bugzilla.samba.org/show_bug.cgi?id=13253


cryptomilk:
@abbra I would be fine with that patch. Can you please take a look?

abbra:
@cryptomilk looks good to me too. Could you please push?

abartlet:
Merged into master as 440ddf8470b11a46066d282bf8945201d547c192 for Samba 4.10

PR #216, opened 2019-02-02
ctdb_mutex_ceph_rados_helper: fix deadlock via lock renewals

QuantumEnergyE:
fix deadlock via lock renewals.
https://bugzilla.samba.org/show_bug.cgi?id=13540


cryptomilk:
I'm sorry, but the Samba 4.6 branch is no longer maintained. See https://wiki.samba.org/index.php/Samba_Release_Planning#General_information

douglasbagnall:
referenced bug https://bugzilla.samba.org/show_bug.cgi?id=13540 is fixed.

PR #222, opened 2023-04-03
fix regression on smbc_getxattr and fix doc

remicollet:
Regression introduce in 113536e0d735a5235f8be29d4fd1cfc8177930b1

This was a documentation issue
The return is the buffer size, as explained at another place

```
 * @param size      The size of the buffer pointed to by value.  This parameter
 *                  may also be zero, in which case the size of the buffer
 *                  required to hold the attribute value will be returned,
 *                  but nothing will be placed into the value buffer.
```

So this is mostly a revert of the change for https://bugzilla.samba.org/show_bug.cgi?id=14808
and fixing the documentation.

Needed in 4.16 and 4.17

### DIFF
--- a/source3/include/libsmbclient.h
+++ b/source3/include/libsmbclient.h
@@ -2419,7 +2419,7 @@ int smbc_getxattr(const char *url,
  *                  required to hold the attribute value will be returned,
  *                  but nothing will be placed into the value buffer.
  *
- * @return          0 on success, < 0 on error with errno set:
+ * @return          size on success, < 0 on error with errno set:
  *                  - EINVAL  The client library is not properly initialized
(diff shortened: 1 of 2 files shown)


anoopcs9:
Please report bugs on [Samba bugzilla](https://bugzilla.samba.org/).

General questions can be directed to our [mailing list](https://lists.samba.org/mailman/listinfo/samba) or even more technical aspects to [samba-techincal list](https://lists.samba.org/mailman/listinfo/samba-technical).

Changes must be submitted via [GitLab repository](https://gitlab.com/samba-team/samba/-/merge_requests) following the instructions from [wiki](https://wiki.samba.org/index.php/Contribute#Submitting_Patches_via_GitLab).

